PR #13, opened 2018-03-05
Documentation updates and defaults

kakra:
This series of patches adds documentation and changes the default debug level.


atar-axis:
I like that modprobe.d config thing, thanks!

PR #15, opened 2018-03-05
Style fixup

kakra:
Note: Please merge PR #13 first.

This commit adds an indentation rules file and fixes most of the code style issues.


atar-axis:
A line breaks after the return type is a style which is encouraged by GNU:
https://www.gnu.org/prep/standards/html_node/Formatting.html#Formatting
but not by the Kernel Coding Style:
https://www.kernel.org/doc/html/v4.10/process/coding-style.html#functions

You are absolutely right, there is A LOT which does not fit the coding style, but not this one - luckily, because I personally dont like that style :smile:

kakra:
Okay, we can simply change `.indent.pro` to reflect the proper style. I just copy and pasted the recommended kernel style from `man indent`.

atar-axis:
as with 702ace99d54e13efc31d0b496a75c65c11981940 the code is (nearly) compliant to the kernel coding style (expect long lines and trailing statements, i.e. break's on switch-case's)

PR #17, opened 2018-03-06
"sudo tee" is probably a better idiom

kakra:
Instead of running escaped command lines through "sudo -c", better use
the "sudo tee" idiom. This has fewer surprises if something does wrong
with the escaping.


atar-axis:
Perfect!

PR #24, opened 2018-03-16
[NOT FOR MERGE] Modified directional rumble test

kakra:
I modified the rumble test by playing the effects in shorter intervals
and repeating a complete circle multiple times.

It looks like the effect is gone now, I cannot really feel how it walks
from left to right.

Also, directional rumble in Wine joy.cpl doesn't work. But I did not
check if it is not implemented there or uses a different interface.

### DIFF
--- a/tools/directional_rumble_test/main.c
+++ b/tools/directional_rumble_test/main.c
@@ -10,6 +10,7 @@
 
 
 enum {
+    FF_DIR_SOUTH      = 0x0000,
     FF_DIR_WEST       = 0x4000,
     FF_DIR_WEST_NORTH = 0x6000,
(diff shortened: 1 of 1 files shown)


kakra:
If you want to test on Wine yourself, you should add my patches from https://github.com/kakra/wine/commits/patches-3.3/xboxcontroller

kakra:
It seems to play rumble on the triggers only the first round.

atar-axis:
Does it work for you the way I mentioned above?

kakra:
I still had no time to test this, I'll do soon.

kakra:
I see you merged my idea of changing the rumble test, so I close this.

PR #31, opened 2018-05-21
Document various caveats

### LINKED ISSUE #27: SDL update adjusts Xbox One S Android mapping
With changeset http://hg.libsdl.org/SDL/rev/6aade58208aa libSDL adjusted the mapping to correctly map the original Android mapping of the controller which renders xpadneo again unusable with games using libSDL.

The most simple solution currently is to export the xpadneo mapping into your environment:

````
export SDL_GAMECONTROLLERCONFIG="\
050000005e040000fd02000003090000,Xbox One Wireless Controller,\
a:b0,b:b1,back:b6,\
dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b8,\
leftshoulder:b4,leftstick:b9,lefttrigger:a2,leftx:a0,lefty:a1,\
rightshoulder:b5,rightstick:b10,righttrigger:a5,rightx:a3,righty:a4,\
start:b7,x:b2,y:b3,"
````

My current solution is to add this mapping to both my Steam installations (native and Wine) in `$STEAM_INSTALL_DIR/config/config.vdf`. It works in Wine only when Wine was compiled with SDL support but without you usually don't have support for the controller in xinput games anyways.

Could there be a better (more automatic) solution?

Is it possible that using xpadneo could result in a unique controller ID for SDL? I'm not sure what it is build from but it consists of four 32-bit words in LSB order:

    unknown1(05000000)+vendor(5e040000)+product(fd020000)+unknown2(03090000)

xpadneo should probably not change the vendor+product part but maybe it could change what SDL sees for the both unknown values? E.g., SDL maps joydev/xinput devices to `unknown1(03000000)` it seems.

Reverting exposed button mapping in xpadneo seems no option as then `/dev/input/js0` will be broken again.

As a start, we could add the needed env var in the README.

**Update:** It seems to be `bustype+vendor+product+version`, see http://hg.libsdl.org/SDL/file/f162f5d7798d/src/joystick/linux/SDL_sysjoystick.c#l226

Maybe version could be faked from the xpadneo driver?

kakra:
When using xpadneo, there a various bad interactions with existing
fixups trying to fix the same thing at the wrong layer. Let's explain
and document those.

Closes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,71 @@ files in `/sys/module/hid_xpadneo/parameters`:
 - `dpad_to_buttons`: Set to Y to map dpad directional input to button events,
   defaults to N
 
+## Caveats
+
+From libSDL 3.3 onward, SDL contains a fix for the wrong mapping introduced
+by the generic hid driver. Thus, you may experience wrong button mappings
+again despite using xpadneo as the driver.
+
+Also, Wine since version 3.3 supports using SDL for `xinput*.dll`, and with
+version 3.4 it includes a patch to detect the Xbox One S controller. Games
+running in Wine and using xinput may thus also see wrong mappings.
+
+The Steam client includes a correction for SDL based games since some
+version, not depending on the SDL version. It provides a custom SDL
+mapping the same way we are describing here.
+
+To fix this and have both SDL-based software and software using the legacy
+joystick interface using correct button mapping, you need to export an
+environment variable which then overrides default behavior:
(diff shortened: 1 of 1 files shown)


atar-axis:
Perfect, thank you!

kakra:
Please pull again, there's a small fix to it.

PR #68, opened 2019-01-04
Document how to get xpadneo to work with Secure Boot

CommunicationAnimale:
In this PR, I document how to make xpadneo and Secure Boot get along.
I added a small note to the README.md, and a longer description to a SECUREBOOT.md file which I included in index.md.
I also added the three helper scripts to the misc/ folder. They are modified so that signing is fully automatic (i.e. passwordless).

I wanted to make some modifications to DEBUGGING.md and bug_report.md in another PR, but since I already committed them to master I did not find how to split it in another PR...
My gripe was that 
- `head -1 "/sys/kernel/debug/hid/0005:045E:<TAB>/rdesc" | tee /dev/tty | cksum` could not be copy-pasted even when I ran it directly as root like you suggested. I think a command using `find` will be more robust.

Finally, I said that Secure Boot could not be disabled on ARM devices, but I found out that Secure Boot could always be disabled at the shim level using `mokutil --disable-validation`. This makes signing instructions nice-to-have but quite secondary.

Hope this helps ;)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -17,10 +17,7 @@ and add at least the following outputs
 - `dmesg`
 
 The following command need superuser privileges.
-I recommend to run it as root directly (do not use `sudo`),
-otherwise the auto-completion using <TAB> will not work as expected.
-
-- `head -1 "/sys/kernel/debug/hid/0005:045E:<TAB>/rdesc" | tee /dev/tty | cksum`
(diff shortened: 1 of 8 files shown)


atar-axis:
Perfect, thank you very much!

CommunicationAnimale:
I forced pushed to fix a typo: a pesky ` slipped in DEBUGGING.md!

CommunicationAnimale:
Well that was fast ;)

CommunicationAnimale:
I think there is a formatting bug on the documentation website: the section "Third party Bugs" ends up in the list of "Notes" from the previous section. Maybe adding a newline at the end of SECUREBOOT.md would fix the issue (I cannot test it on my machine).

atar-axis:
Ah yes I will fix it, thank you for pointing out!

atar-axis:
Done.

PR #82, opened 2019-03-13
Added configure.sh and accompanying doc

CodeCanna:
Ready for review and testing.


atar-axis:
will give it a try tonight, thank you very much so far :)

atar-axis:
The script is not able to detect my (installed, but not active) xpadneo instance:
```
flood@flood-r3:[xpadneo]$ sudo ./configure.sh 
[sudo] Passwort für flood: 
Installaion not detected, did you run install.sh?
```

I guess this is because I had an older version installed - therefore you should also check for older versions and complain if the version does not fit the configure script :)

Btw, there is a typo: "Installaion"

Another thing is, that we permanently chang the options using modprobe, as described in the README - this does also work if the module is not loaded :)

CodeCanna:
Should the script run if a different version is detected?

atar-axis:
No, I think it would be enough to report an error - otherwise we would have to check if a given parameter is available in the installed version. That's too much I think :)

CodeCanna:
Okay, so now the script checks if the version is the same is the number in the VERSION file in the xpadneo directory, and throws a warning before exiting.

atar-axis:
Perfect, the only thing missing now is the permanent change in /etc/modprobe.d/xpadneo.conf 😊

CodeCanna:
So just to be sure, in order to make changes permanent you put ```options hid_xpadneo debug_level=3``` in xpadneo.conf.  If I want to do disable_ff I would put ```options hid_xpadneo disable_ff=true```  and so on, is it all the same for each option?

atar-axis:
I hope this helps: https://wiki.archlinux.org/index.php/Kernel_module#Setting_module_options

atar-axis:
To take it short: yes! The only, thing you need to take care about is that you do not overwrite existing options

CodeCanna:
I'm a little confused about parsing the conf file.  I was thinking of using ```echo``` to put the appropriate line into the conf like:

```echo "options hid_xpadneo debug_level=2"  >>  /etc/modprobe.d/99-xpadneo-bluetooth.conf```

But I think that would cause duplicate lines if that option was changed more than once.  Is that why you use tee in the example?


> will give it a try tonight, thank you very much so far :)
> btw, codacy complains about an unused ARGS variable, it it a bug or is it codacy which is wrong?

Also sorry, I didn't see the rest of this post, I didn't see where you asked about the variable.  It was the ```$ARGS``` variable which didn't end up being used.

**Comment by CodeCanna** [comment]
Ah looks like I can use ```sed``` for this.  A new command for me.

**Comment by CodeCanna** [comment]
I need an opinion, should I add all of the config lines in ```/etc/modprobe.d/xpadneo.conf``` when the script runs.  Because the only line that is in there by default is ```options bluetooth disable_ertm=y```, and ```sed``` won't edit the line if it doens't exist.  So I have come up with either echoing all the config lines into the file when the script is ran, or using ```grep``` to check if the line exists or not, and if not to create it so ```sed``` can edit it.

**Comment by atar-axis** [comment]
> It means a LOT to me

I can remember that time :) 
I am happy that it is fun to you too

> Ah looks like I can use sed for this

Right! `sed` is the way to go I think, or perl - which regex syntax is a bit easier. Parsing the config file is maybe the Most difficult part of this Script

> I need an opinion

`options bluetooth ...` has nothing to do with the xpadneo configs, it's just in the same file.
I think thats clear, right?


The latter option sounds better to me - in the end it is up to you, as long as the solution is robust and clean ;)

**Comment by CodeCanna** [comment]
> `options bluetooth ...` has nothing to do with the xpadneo configs, it's just in the same file.
> I think thats clear, right?

No the file has that line in it by default.  When I opened the config file that line was already in there.  Is it supposed to be empty?

**Comment by atar-axis** [comment]
Everything is alright, this Line is in there to disable the faulty ERTM in Linux which prevents a sucessful connection with the gamepad 😊

**Comment by CodeCanna** [comment]
I might have found a better aproach to using bash to parse the config file.  I have seen that Python has a good [library](https://docs.python.org/3/library/configparser.html) for this.  What do you think?

**Comment by atar-axis** [comment]
I don't like the Idea to add another dependency for this little script - not everybody has Python installed. If you don't like Shell scripts, then I would suggest perl, which is installed in most (all?) Distributions. 😊

**Comment by CodeCanna** [comment]
Hmm never thought about perl.  Is it a big language in linux?  I'll look into it.  If that doesn't work and time isn't an issue I will figure something out.  If time is an issue I have a method that will work, it will just be redundant.  I'm just really trying to use a function instead of writing the same code for each option to parse the config file.

**Comment by atar-axis** [comment]
Time is not a problem - don't worry, that's the good thing in open source projects like this 😊

**Comment by CodeCanna** [comment]
Awesome, I felt like I was taking too long.  I'm glad, now I can take time to really learn some of these concepts. :relieved:

**Comment by atar-axis** [comment]
I opened this issue *two month* ago, take your time ;)

**Comment by CodeCanna** [comment]
Alright!!  Give that a try, I rewrote the entire thing.  I discovered this awesome handy tool that everyone else probably knows about already called getopt.  Made it soooo much easier and cleaner.  And I stuck everything into functions this time.  I had a good time writing this one!  :smiley:

**Comment by atar-axis** [comment]
Congratulations! It looks very clear now!

If the module is not loaded, then the script returns the following:
```
flood@flood-r3:[xpadneo]$ sudo ./configure.sh -d 3
./configure.sh: Zeile 142: /sys/module/hid_xpadneo/parameters/debug_level: Datei oder Verzeichnis nicht gefunden
./configure.sh:Problem writing to /sys/module/hid_xpadneo/parameters
```

It is sufficient to modify the modprobe *.conf* if the module is not loaded. Modprobe will use the configuration in the .conf file next time the module gets loaded - no need to change the configs *on the fly* using the sysfs files in `/sys/module/...`.

Therefore you should probably check if the module is loaded or not :grin:

**Comment by CodeCanna** [comment]
SWEEEET!  I fixed the module detection problem.  It should be ready to roll! :smile:

**Comment by atar-axis** [comment]
:grin: the output now is:

```
flood@flood-r3:[xpadneo]$ sudo ./configure.sh 
[sudo] Passwort für flood: 
./configure.sh:Module not detected, try running 'modprobe hid_xpadneo' and try again.
```
Instead, I would - as explained above - prefer to let the module stay unloaded and simply do not edit the files in `/sys/module/...`. 

Next time the module gets loaded, it gets loaded with the configuration in `xpadneo.conf`, therefore it is not necessary to load the module just to configure it ;)

**Comment by CodeCanna** [comment]
Ohhhhhhhhh, I didn't get what you were saying.  Ok so the module doesn't have to be loaded to edit the conf in /etc/modprobe.d?

**Comment by atar-axis** [comment]
right! :grin: the `xpadneo.conf` file the `modprobe` directory defines *default* settings, which are set at the moment the module gets loaded! setting the options in `/sys/module/...` takes effekt immediately, but only until the module gets unloaded ;)

**Comment by CodeCanna** [comment]
so if you change a setting using this script, you have to reload the module?  And also do you still want to check for the /sys/src installation?

**Comment by atar-axis** [comment]
I edited my answer, hope it is a bit clearer now :)

**Comment by CodeCanna** [comment]
> Therefore you should probably check if the module is loaded or not
 
if the module IS loaded, what do I do vs. if the module is NOT loaded.

> Instead, I would - as explained above - prefer to let the module stay unloaded and simply do not edit the files in

So are we letting loading the module up to the user?  What if the user has the module inserted when the script is ran?

> takes effekt immediately, but only until the module gets unloaded

The config only takes effect upon unloading if the module is current inserted and the settings are changed?

Bear with me once I understand, I'll get it for good it just takes me a lil bit :)

**Comment by atar-axis** [comment]
If the module is unloaded, you only need to modify `xpadneo.conf'
If the module is loaded, you need to modify `/sys/modules/.../parameters` AND `xpadneo.conf'

The reason is the following:

If a module gets loaded, then modprobe is searching in those `.conf` files for fitting `options`. If there is a config file that fits the module (`xpadneo.conf` in this case), then it loads the module with the settings defined in this config file.
This is done at loading time, and only then!

You can than see those option in `/sys/modules/.../parameters` once the module is loaded. The files in this directory reflect the **current state** of the module . The cool thing is: you can also change those settings **on the fly** (while the module is loaded) - those changes will take immediately and last until the module gets unloaded.

Next time the module gets loaded it will - once again - be loaded with the **default** settings in `xpadneo.conf'.

**Comment by CodeCanna** [comment]
Whenever I try to load the module again after being unloaded I get this error:

```modprobe: ERROR: could not insert 'hid_xpadneo': Invalid argument```

and I have to run the uninstall script and the install script to be able to insert it again.  Am I doing something wrong?  I'm using ```modprobe -r hid_xpadneo``` to remove the module and ```modprobe hid_xpadneo``` to insert it.  I think I got it though, it should only now edit parameters if the module is inserted as well as the conf, and should only edit the conf if its not inserted.

**Comment by atar-axis** [comment]
can you give me the content of your `xpadneo.conf` and the output of `dmesg` (just the last couple of lines) when that error occurs please?

**Comment by CodeCanna** [comment]
I might have found the problem, now the script is copying lines instead of checking if its already there.  So I have 3 debug_level lines in my config file.

Config:
```
options bluetooth disable_ertm=y
options hid_xpadneo debug_level=2
options hid_xpadneo debug_level=1
options hid_xpadneo debug_level=0
options hid_xpadneo disable_ff=true
```
Dmesg:
```
[  645.922742] hid_xpadneo: `false' invalid for parameter `disable_ff'
[  645.922744] hid_xpadneo: `true' invalid for parameter `combined_z_axis'
[ 4610.952246] hid_xpadneo: `true' invalid for parameter `disable_ff'
[ 4952.497307] hid_xpadneo: `true' invalid for parameter `disable_ff'
```

**Comment by CodeCanna** [comment]
It's because I changed my cat statements, and it can't tell if the line exists anymore so it just echos.

Original:
```
local LINE_EXISTS=$(cat $CONF_PATH | grep debug_level)
```

New:
```
local LINE_EXISTS=$(grep -q "debug_level" "$CONF_PATH")
```

**Comment by CodeCanna** [comment]
Never mind it's still giving me an error when I try to re-insert the module.  Same error message.  Sorry I thought it was the copying of lines in config, but I have fixed that.

Config:
```
options bluetooth disable_ertm=y
options hid_xpadneo debug_level=1
options hid_xpadneo disable_ff=true
```

dmesg:
```
[ 5997.480466] xpadneo: goodbye!
[ 6001.641547] xpadneo: hello there!
[ 6616.654502] logitech-hidpp-device 0003:046D:101B.0007: multiplier = 8
[ 6626.936481] logitech-hidpp-device 0003:046D:101B.0007: multiplier = 8
[ 6804.611256] xpadneo: goodbye!
[ 6821.624267] hid_xpadneo: `true' invalid for parameter `disable_ff'
```

**Comment by atar-axis** [comment]
`[ 6821.624267] hid_xpadneo: `true' invalid for parameter `disable_ff'`

try `y` / `n` or `1` / `0`

**Comment by CodeCanna** [comment]
Ok, I changed the booleans from true/false to y/n and it seems to work.  See if it works from your end.  I have some other ideas for your driver as well.  I guess I open up another issue on here?  But I'll wait to see if this script works on your end first.

**Comment by atar-axis** [comment]
1) ```
flood@flood-r3:[xpadneo]$ sudo ./configure.sh -d 0
grep: xpadneo.conf: Datei oder Verzeichnis nicht gefunden
./configure.sh:Debug Level set to 0.
./configure.sh:Config written to /etc/modprobe.d/99-xpadneo-bluetooth.conf
xpadneo.conf
```

2) seems like the script created a file in:
``` 
/etc/modprobe.d/99-xpadneo-bluetooth.conf\nxpadneo.conf
```

atar-axis:
Hint: I guess you are missing the `shift` command ;)
Take a look here: https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash

CodeCanna:
I have no clue how to make it take multiple arguments at once, that's something I'm still learning.  For now I guess it just takes one at time.  When I learn how to fix that I'll add it, that's one of the things I'm still learning.

>Please change the description from [true, false] to [y, n]

If your talking about the help file, it is changed from true false to y, n.  Guess I forgot to push that file.

>seems like the script also created a file

Thats super weird, I have no clue why it does that.  In my system that is the name of the xpadneo config file, not xpadneo.conf.

atar-axis:
> If your talking about the help file, it is changed from true false to y, n. Guess I forgot to push that file.

Yes, that's what I was talking about ;)

> Thats super weird

It doesn't matter - I don't think that we have to cover this

> I have no clue how to make it take multiple arguments at once

See the stackoverflow answer in my last post

CodeCanna:
But if it creates a directory that is the same name as the config file I have on my system, then how will the driver find it?  That seems pretty important, I use grep to hopefully catch any folder with xpadneo in the name.

atar-axis:
As said, it was just a problem because I had my old version installed. But we will change the version number anyway if we merge your pull request, therefore this cannot happen then :)

CodeCanna:
Did the version check not work?  I put a function in there that is supposed to exit if the version numbers are different.

atar-axis:
the version was the same I guess! since we haven't yet increased the number in your merge request.
I found the problem, I had two different config files in my modprobe.d directory!
* 99-xpadneo-bluetooth.conf
* xpadneo.conf

CodeCanna:
Alright!  It's ready for another try, I am able to pass multiple arguments and it seems to work.  It was because I had a bunch of un-needed ```exit 0``` statements after each option completed successfully and I didn't have a ```--``` case.  It's actually a lot simpler than it looked but learning the concept of ```shift``` was a little daunting.

atar-axis:
Wohoo! Looks like it ready now :)
Two last things:

1) Did you know that you don't need `options hid_xpadneo` on every option?
Instead of:
```
options hid_xpadneo debug_level=3
options hid_xpadneo combined_z_axis=y
```
you can write
```
options hid_xpadneo debug_level=3 combined_z_axis=y
```
but I think that complicates parsing, right?

2) Would you please increase the Version number by one? (It is in VERSION)

CodeCanna:
Nice!  I didnt know you could put the options on one line.  I'll mess around with that.

I'll increase the version number when I get back.

As far as malicious cases I learned a lot about the '''$?''' variable.  I helps a lot with detecting successful writes or unsuccessful. But hopefully the users will be able to break it and bend it and find those I didnt cover.  😀

CodeCanna:
Alright version number is now 0.5.5.

CommunicationAnimale:
Hi, I stumbled upon your script ; here are some comments that you may find useful:
* It is generally considerde [bad practise to parse `ls`](https://mywiki.wooledge.org/ParsingLs) (or [stack overflow](https://unix.stackexchange.com/questions/128985/why-not-parse-ls-and-what-do-to-instead). Instead, you could use the utility `find`:
```console
$ find /usr/src -mindepth 1 -maxdepth 1 -type d -name "hid-xpadneo*"
/usr/src/hid-xpadneo-0.5.4
```
Thus you could use:
```sh
SOURCE_PATH=$(find /usr/src -mindepth 1 -maxdepth 1 -type d -name "hid-xpadneo*")
```
(Though honestly it only fails in really special cases)

* To get the version number, instead of listing the characters that you don"t want to keep, you could use `sed` and list the characters that you do want to keep:
```console
$ echo '/usr/src/hid-xpadneo-0.5.4' | sed "s/[^[:digit:].]//g"
0.5.4
```
(`^` means 'except', `g` means 'Apply the replacement to all matches to the regexp, not just the first. '. The brackets create a 'character class')

Better yet: keep only dots and digits at the end of the filename (i.e. after the last dash):
``` console
$ echo "/usr/src/hid_xpadneo-0.5.4" | sed -E 's/.*-([[:digit:].]+)/\1/'
0.5.4
```
(I used `-E` to enable the extended regex syntax: this way you can create 'capture groups' (in parenthesis) without having to escape the parenthesis. I can then refer to the capture group using `\1`. Here the regex can be read this way: match everything until a dash followed by dots and digits, and return only these dots and digits.)

* To parse lines such as
```
options hid_xpadneo debug_level=3 combined_z_axis=y
```
you may also use capture groups:
```console
$ CFG="options hid_xpadneo debug_level=3 combined_z_axis=ysed"  
$ echo "$CFG" | sed -E "s/(^\s*options\s*hid_xpadneo.*)debug_level=\S*(.*)$/\1debug_level=100\2/"
options hid_xpadneo debug_level=100 combined_z_axis=ysed
```
(`\s` is a shorthand for 'whitespace character' and `\S` is a shorthand for 'non-whitespace character')

You could even abstract it to a more general function:
```sh
function set_option () {
  key=$1
  value=$2
  sed -i'' -E \
    "s/(^\s*options\s*hid_xpadneo.*)$key=\S*(.*)$/\1$key=$value\2/" \
    "$CONF_PATH"
}
```

I ran `set_option debug_level 100` on
```
options hid_xpadneo combined_z_axis=y
options hid_xpadneo atest=5 debug_level=3 btest=4
  options hid_xpadneo ctest=5   debug_level=6 dtest=4
options hid_xpadneo combined_z_axis=n
```
and got back 
```
options hid_xpadneo combined_z_axis=y
options hid_xpadneo atest=5 debug_level=100 btest=4
  options hid_xpadneo ctest=5   debug_level=100 dtest=4
options hid_xpadneo combined_z_axis=n
```
Nice :)

CodeCanna:
Thanks so much for the great info.  I'm glad to take advice, because I NEED it.

>Thus you could use:
>```SOURCE_PATH=$(find /usr/src -mindepth 1 -maxdepth 1 -type d -name "hid-xpadneo*")```

Does this mean I can get rid of ```$XPAD_DIR```?  I figured that was a redundant variable :)

>To get the version number, instead of listing the characters that you don"t want to keep, you could use ```sed``` and list the characters that you do want to keep:

I didn't know `sed` was that powerful, then again I'm still learning how to use it.

>I ran set_option debug_level 100 on
>```
>options hid_xpadneo combined_z_axis=y
>options hid_xpadneo atest=5 debug_level=3 btest=4
>  options hid_xpadneo ctest=5   debug_level=6 dtest=4
>options hid_xpadneo combined_z_axis=n
>```
>and got back
>```
>options hid_xpadneo combined_z_axis=y
>options hid_xpadneo atest=5 debug_level=100 btest=4
>  options hid_xpadneo ctest=5   debug_level=100 dtest=4
>options hid_xpadneo combined_z_axis=n
>```

I didn't know you can have 2 of the same option, in this case which one of the `debug_level` options is being actually used, and I'm assuming that you didn't want both of the values changed right?

>You could even abstract it to a more general function:

>function set_option () {
  key=$1
  value=$2
  sed -i'' -E \
    "s/(^\s*options\s*hid_xpadneo.*)$key=\S*(.*)$/\1$key=$value\2/" \
    "$CONF_PATH"
}

This is beautiful, I was really trying to make a function in the first version I wrote that changed the options like that, but I still had yet to discover `sed`.  Thank you so much for the words of wisdom.  I'm gonna go start messing with this, and will hopefully have a better version.

CommunicationAnimale:
> Thanks so much for the great info. I'm glad to take advice, because I NEED it.

I'm glad you appreciate :)

> Does this mean I can get rid of `$XPAD_DIR`? I figured that was a redundant variable :)

Yep, I guess so.


> I didn't know you can have 2 of the same option, in this case which one of the `debug_level` options is being actually used, and I'm assuming that you didn't want both of the values changed right?

Actually I don't know: normally you will NEVER have the same options specified multiple times. But sometimes things don't go as planned... I don't really know what to do in such a case.

> This is beautiful, I was really trying to make a function in the first version I wrote that changed the options like that, but I still had yet to discover `sed`. Thank you so much for the words of wisdom. I'm gonna go start messing with this, and will hopefully have a better version.
> 
> EDIT:
> 
> > `echo "$CFG" | sed -E "s/(^\s*options\s*hid_xpadneo.*)debug_level=\S*(.*)$/\1debug_level=100\2/"`
> 
> This line is using the `\s` in front of the "options" and "hid_xpadneo", which means that there can be anything before it but we just want that line with options? What are all of the `\s` and `\l` called? Are they delimiters?

Not anything, only whitespace: I guess a file like this is valid
```
                   options hid_xpadneo debug_level=3
```
so I tried to deal with it.
\s and \S are actually shorthand for character classes:
* `\s` = `[[:space:]]` = whitespace characters
* `\S` = `[^[:space:]]` = non whitespace characters.
Therefore `\s` needs to match exactly one whitespace character, and `\s*` will match zero, one, or more whitespace characters (sed regexes are calles 'greedy': patterns will try to match the largest string possible. Therefore `^\s*` will match all whitespace characters at the beginning of the line.)

I see you have enabled the POSIX extension: `\S` and `\s` are actually not POSIX compliant (they are GNU extensions to sed). If you wish to remain POSIX compliant you should use `[[:space:]]` and `[^[:space:]]`.

Good luck going down the rabbit hole of regular expressions ;)

CommunicationAnimale:
I have just read the part ["Sed overview"](https://www.gnu.org/software/sed/manual/sed.html#sed-script-overview) in the GNU Sed manual, and I have found an invaluable feature: you can ask sed to only work on lines matching a certain regex (read the part about "the optional line address [addr]"). For instance
```
sed "/foo/s/bar/baz/"
```
will replace the first instance of `bar` by `baz` but only on lines containing `foo`. If we add the `g` flag
 lines matching a certain regex (read the part about "the optional line address [addr]"). For instance
```
sed "/foo/s/bar/baz/g"
```
it will replace every instance of `bar` with `baz` but only on lines containing `foo`.

We may now rewrite our function like this:
```sh
function set_option () {
  key=$1
  value=$2
  sed -i'' \
    "/^[[:space:]]*options[[:space:]]*hid_xpadneo/s/$key=[^[:space:]]*/$key=$value/g" \
    "$CONF_PATH"
}
```
This will replace EVERY instance of `$key=...` with `$key=$value` but only on lines beginning by `options hid_xpadneo` (+ eventual whitespace).

CommunicationAnimale:
Now when I run the function on
```
options hid_xpadneo combined_z_axis=y
options hid_xpadneo atest=5 debug_level=8 btest=4
useless line
  options hid_xpadneo debug_level=4 debug_level=6 dtest=4
options hid_xpadneo combined_z_axis=n
```
I get
```
options hid_xpadneo combined_z_axis=y
options hid_xpadneo atest=5 debug_level=100 btest=4
useless line
  options hid_xpadneo debug_level=100 debug_level=100 dtest=4
options hid_xpadneo combined_z_axis=n
```
I think this is a way more predictable behavior.

CodeCanna:
>\s = [[:space:]] = whitespace characters
\S = [^[:space:]] = non whitespace characters.

So is `s/` and `s\` the same thing?

CommunicationAnimale:
> > \s = [[:space:]] = whitespace characters
> > \S = [^[:space:]] = non whitespace characters.
> 
> So is `s/` and `s\` the same thing?

No:
* s/ means 'do a substitution': sed expressions can do many things (see https://www.gnu.org/software/sed/manual/sed.html#sed-commands-list for a list of every available command), one of them is substitution. The syntax for a substitution is `s/REGEXP/REPLACEMENT/[FLAGS]` (actually the delimiters can be anything so `s!REGEXP!REPLACEMENT![FLAGS]` and `s@REGEXP@REPLACEMENT@[FLAGS]` are also valid)
* \s represents the character class `[[:space:]]` in a REGEXP.
```
sed "s@aaa\sbbb@def@" input
````
is the same thing as
```
sed "s/aaa[[:space:]]bbb/def/" input
```
and means "Replace the first occurrence of `aaa bbb` or `aaa<TAB>bbb` with `def`".

**Comment by CodeCanna** [comment]
I hope ataris-axis is ok, I wonder where he is?

**Comment by atar-axis** [comment]
For sure I am :) Just giving you some space to work on it without hurry. If you have any questions - just ask. Otherwise give me a kick when you are done so - I will then test it :grin:

**Comment by CodeCanna** [comment]
Oh, ok cool!  Glad everything is good!  I'm just diving into the rabbit hole of regexes like CommunicationAnimale said, try to get a good grasp on the set_option function that he wrote for me.

**Comment by CodeCanna** [comment]
So I have been spending the last couple of days learning about regexes and they are making me want to pull my hair out.  The regex in `set_option()` is not working for me, I have been trying to fix it but regexes are gonna be a crazy beast to tame.  This is like when I first was told what piping was, it took a lil while but once I got it, it stuck.

This regex: `sed -i'' "/^[[:space:]]*options[[:space:]]*hid_xpadneo/s/$key=[^[:space:]]*/$key=$value/g" \
    "$CONF_PATH"` isn't working for me.  I have spent time trying to figure out why, but I think i need some help figuring this out.  It simply doesn't change anything in the conf file.

**Comment by atar-axis** [comment]
the following replaces all "testkey=" values with "testkey=newvalue" in the test.conf file
```
#!/bin/bash
key=testkey
value=newvalue
sed --in-place --debug "/^[[:space:]]*options[[:space:]]*hid_xpadneo/s/$key=[^[:space:]]*/$key=$value/g" test.conf`
```

test.conf before running the script:
```
options hid_xpadneo testkey=change testkey=me
testkey=shouldnotchange
```

test.conf after:
```
options hid_xpadneo testkey=new testkey=new
testkey=shouldnotchange
```

**Comment by atar-axis** [comment]
I'll try to explain the SED command:

`sed --in-place "/^[[:space:]]*options[[:space:]]*hid_xpadneo/ s/$key=[^[:space:]]*/$key=$value/g" test.conf`

It contains of two different commands, the first one is:
`/^[[:space:]]*options[[:space:]]*hid_xpadneo/`
it is explained here: https://www.gnu.org/software/sed/manual/html_node/Regexp-Addresses.html#Regexp-Addresses

What it does is, that it selects only those lines which start with `options hid_xpadneo` including any count of whitespaces. I, personally, think that 
`/^[[:space:]]*options[[:space:]]+hid_xpadneo/`
would be a better choice because between `options` and `hid_xpadneo` should be at least one whitespace.

the second command is `s/$key=[^[:space:]]*/$key=$value/g`, so in every line selected by the first command, every string that matches `$key=[^[:space:]]*` is replaced by `$key=$value`, this is done as often as possible by the `g` modifier.

I like extended regexp with back-references and so on, that makes life a lot easier. Just as a side note.

**Comment by atar-axis** [comment]
so what exactly doesn't work on your system? can you please share your code with us? one thing we should mention is that you are not allowed to add any whitespace after a line-continuation (`\`)

**Comment by CodeCanna** [comment]
aaaaannnnddddd I'm an idiot.  In my test file, I WASN'T CALLING THE FUNCTION!!!!!!  How STUPID!! *bangs head with keyboard for 10 minutes*

I wrote a separate script to test the function and mess around with it.

Sorry about that, apparently I had a mental break down.  Remember kids CALL YOUR FUNCTIONS WHEN YOU WRITE AND INTEND TO USE THEM!!

Again sorry though for making a big deal about this, it was my error not your guy's.

Wow, well here is the code, pretty much copied and pasted from CommunicationAnimale's post, which works if you actually call the function by the way.

```
CONF_PATH=./test.conf

function set_option () {
    key=$1
    value=$2
    sed -i'' "/^[[:space:]]*options[[:space:]]*hid_xpadneo/s/$key=[^[:space:]]*/$key=$value/g" "$CONF_PATH"
}

set_option $@  # Here is the function i forgot to call, YAY!!
```

**Comment by atar-axis** [comment]
As you already know, in regexp:
`*` means **zero or more**
`\+` means **one or more**

Therefore I would suggest changing the regexp to
<pre><code>
"/^[[:space:]]*options[[:space:]]<b>\+</b>hid_xpadneo/s/$key=[^[:space:]]*/$key=$value/g"
</code></pre>

Otherwise, `optionshid_xpadneo` would match too.

Furthermore:
1. You don't need the quotes here: `"$CONF_PATH"`. There are two types of quotes in shell scripts, the first one is `'...'`, the second one is `"..."`. The difference is, that in `'...'` variable names (like $CONF_PATH) aren't evaluated, in `"..."` they are. This is useful e.g. in rexep expressions, but you don't need them if it is only the variable itself.
You can see the difference here:
```
#!/bin/bash
CONF_PATH=foo.conf
echo 'variable name in single quotes: $CONF_PATH'
echo "variable name in double quotes: $CONF_PATH"
```
the output is:
```
variable name in single quotes: $CONF_PATH
variable name in double quotes: foo.conf
```

2. you dont need the `''` after sed -i, in fact, I don't know the reason why @CommunicationAnimale added them.

3. I prefer to use long names in scripts, so instead of `-i` I would always use `--in-place` in scripts, makes it easier for others to understand the code

**Comment by CommunicationAnimale** [comment]
> Some suggestions:
>     0.  About `*` and `+`

Absolutely right!

>  1. You don't need the quotes here: `"$CONF_PATH"`...

I don't think this is correct in the general case: if the name of the configuration files contains whitespace or special characters, quotes (these -> `"`) will be needed. `shellcheck` complains if this variable is not quoted:

```console
sh-4.4$ cat quotes.sh
#!/bin/bash
CONF_PATH=$(find . -maxdepth 1 -mindepth 1 -name test)

sed -e "s/a/b/" $CONF_PATH

sed -e "s/a/b/" "$CONF_PATH"
sh-4.4$ shellcheck quotes.sh

In quotes.sh line 4:
sed -e "s/a/b/" $CONF_PATH
                ^--------^ SC2086: Double quote to prevent globbing and word splitting.

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

>  3. you dont need the `''` after `sed -i`, in fact, I don't know the reason why @CommunicationAnimale added them.

I think using `-i''` is a good habit to take: BSD sed has no long options (`--in-place`), and sed on mac requires the suffix to be specified (like this `-i ""`), and finally GNU sed does not allow whitespace between the `-i` and the empty quotes so for full compatibility with the three implementations only `-i''` is possible. (Nonetheless it does not matter here since the script will only run on GNU/Linux)

atar-axis:
> I don't think this is correct in the general case: if the name of the configuration files contains whitespace or special characters

Good point! You are right, for sure.

> I think using -i'' is a good habit to take

I see! Another good point :) Thank you for clarification ;)

atar-axis:
From now on, every PR is automatically checked with shellcheck (via code climate)! Thank you ;)

CodeCanna:
Okay, now one thing I'm still wondering about.  Am I correct about checking for the existence of the line `options hid_xpadneo` with `grep` before trying to use sed to edit the line; and if we are going with one line of `options hid_xpadneo` with the options appended to that line, should I just check for that line once instead of in each function? If it doesn't exist, echo it, then append the options? 

That way I shouldn't have to check for the existence of that line in every function.  In the vanilla config file after a fresh install there are no lines that have `options hid_xpadneo` so I'm wondering if the below code is necessary:
```
  if [[ $LINE_EXISTS ]];
  then
    sed -i 's/options hid_xpadneo debug_level=.*/options hid_xpadneo debug_level='"$value"'/' $CONF_FILE
    if [[ $? -ne 0 ]];  # Did above command run successfully?
    then
      echo "There was a problem writing to $CONF_FILE"
      exit 1
    else
      echo $NAME:"Debug Level set to $value."
      echo $NAME:"Config written to $CONF_FILE"
    fi
  else
    echo "options hid_xpadneo debug_level=$value" >> $CONF_FILE   # Write to config file.
    if [[ $? -ne 0 ]];
    then
      echo "There was a problem writing to $CONF_FILE."
      exit 1
    else
      echo $NAME:"Debug Level set to $value."
      echo $NAME:"Config written to $CONF_FILE"
    fi
fi
``` 
This [post](https://stackoverflow.com/questions/3557037/appending-a-line-to-a-file-only-if-it-does-not-already-exist) seems to agree with `grep` and `echo`, although they are using `grep -qxF 'include "/configs/projectname.conf"' foo.bar || echo 'include "/configs/projectname.conf"' >> foo.bar` instead of assigning `$(grep debug_level $CONF_FILE)` to a variable.

atar-axis:
Am I right that you want to use `grep` before using `sed` to check if there is an `options hid_xpadneo` in the config file? And am I right that you want to use the return code of grep to decide if you need `sed` or echoing is  enough?

If, then you can use the following:
```
#!/bin/bash
key=testkey
value=newvalue
sed --in-place "/^[[:space:]]*options[[:space:]]\+hid_xpadneo/!{q10}; {s/$key=[^[:space:]]*/$key=$value/g}" test.conf
echo "return code of sed was: $?"
```

using this regexp, sed will return `0` if `options hid_xpadneo` was found, and will return `10` if not.

atar-axis:
```
#!/bin/bash
key=testkey
value=newvalue
sed \
-e "/^[[:space:]]*options[[:space:]]*hid_xpadneo/ {s/$key=[^[:space:]]*/$key=$value/g; t bla; b end; :bla h};" \
-e ':end ${x; /./{x; q0}; x; q1}' \
test.conf
echo "return code of sed was: $? (0: substitution execute, 1: otherwise)"
```

atar-axis:
sed is a monster... really, you can do nearly everything with `sed`.
The following checks if a substitution was done or not, if not, it returns exit code 1.

```
#!/bin/bash
key=testkey
value=newvalue
sed \
-e "/^[[:space:]]*options[[:space:]]*hid_xpadneo/ {s/$key=[^[:space:]]*/$key=$value/g; t put_into_hold; b; :put_into_hold h};" \
-e '${x; /./q0; q1}' \
test.conf
echo "return code of sed was: $? (0: substitution execute, 1: otherwise)"
```

the resulting sed script is the following:
```
  /^[[:space:]]*options[[:space:]]*hid_xpadneo/ {
    s/testkey=[^[:space:]]*/testkey=newvalue/g
    t ps_into_hs
    b
    :ps_into_hs
    h
  }
  $ {
    g
    /./ q 0
    q 1
  }
```

Let me explain that a bit:
* you already know that `/^[[:space:]]*options[[:space:]]*hid_xpadneo/` is like a filter for the following command
* the following command is in fact a group of commands:
    ```
    {
        s/testkey=[^[:space:]]*/testkey=newvalue/g
        t ps_into_hs
        b
        :ps_into_hs
        h
      }
    ```
  * you know the first part: `s/$key=[^[:space:]]*/$key=$value/g;` which is the substitution
  * the next command is `t ps_into_hs`, the idea is to jump to the label `ps_into_hs` if (and only if) a substitution occured
  * next comes the `b` which is a branch command and without label, it jumps to the end of the script. remember that this command is not executed when there was a substitution (because of the `t` command)
  * now there is the destination of the `t` command: `:ps_into_hs h`. the `h` command simply puts the PATTERN_SPACE (PS) into the HOLD_SPACE (HS). The PS always contains the current line, the HS is like a personal buffer. So when there was an execution, you use the HS as some kind of flag.
* the last command group is
    ```
  }
  $ {
    g
    /./ q 0
    q 1
  }
    ```
    this is the last command-group, so this is where `b` jumps to. `$` matches only the last line in the file, so whenever the currently readed in line wasn't the last one, the commands in to curly brackets aren't executed. But IF it was the last line, then we execute `g`, which puts our HS again in the PS, we are then looking if there there is anything in it by using `/./`. If there is anything in, then we quit (`q`) and return `0`, otherwise we quit with `1`.

but, this is really hard to maintain and even harder to understand...
just use `grep` :grin:

**Hint:** feel free to use `awk` if you think `sed` you like it better

atar-axis:
Btw, checking once should bei fine

CodeCanna:
>sed is a monster... really, you can do nearly everything with sed.

Sed is reminding me of vim, super powerful but super hard to learn.

atar-axis:
Yeah absolutely! How is it - can I help you somehow?

CodeCanna:
I haven't had time to work on it since Friday, I'm about to get home and I'll type you back when I'm back.  But it's good I got that function working well I just need to add it into the script.

atar-axis:
Just tell me/us if you need some help 😊 Good luck!

CodeCanna:
I appreciate it!  I've been in the mountains so I haven't been doing much coding.  I'm refreshed and hopefully will have a fresh perspective.  I bet this script would have taken a more experienced scripter half a day huh?

That's the level I hope to be soon. :)

CodeCanna:
I do have one question, in the arguments I have them formatted as `--debug-level=4`, but I have ran into a problem with sed not liking the `--` in the options so the `$key=$value` doesn't work with that format.  Would it be easier for me to just take away all of the dashes and replace them with unserscores?  getopt seems to like the `--`, but `sed` doesn't match correctly with the dashes.

ex: `--debug-level=2` vs `debug_level=2`

Also just messing with it now I have ran into a problem where it doesn't recognize the `=`.

Code:
```
#!/bin/bash

if [[ "$EUID" -ne 0 ]];
then
  echo "Script must be ran as root!"
  exit -3
fi

CONF_FILE=./test.conf
LINE_EXISTS=$(grep "options hid_xpadneo" "$CONF_FILE")

if [[ ! "$LINE_EXISTS" ]];
then
  echo "options hid_xpadneo noodle=0" >> "$CONF_FILE"
fi

function set_option {
    key=$1
    value=$2
    sed -i "/^[[:space:]]*options[[:space:]]\+hid_xpadneo/s/$key=[^[:space:]]*/$key=$value/g" "$CONF_FILE"
}

set_option $@
if [[ $? -ne 0 ]];
then
    echo "There was a problem writing to $CONF_FILE"
fi

# The only way this works it by doing ./set_option noodle 3
# This doesn't work: ./set_option --noodle=3
# So I think I'm gonna re-do the argument system to use noodle instead of --noodle
```

atar-axis:
Can you add `set -x` to your script and show us what it outputs? `--` is common practice, please do not change that :)

CodeCanna:
Wow set -x is pretty cool, but anyway here is the ouput of that:
```
+ CONF_FILE=./test.conf
++ grep 'options hid_xpadneo' ./test.conf
+ LINE_EXISTS='options hid_xpadneo noodle=33'
+ [[ ! -n options hid_xpadneo noodle=33 ]]
+ set_option --noodle=300
+ key=--noodle=300
+ value=
+ sed -i '/^[[:space:]]*options[[:space:]]\+hid_xpadneo/s/--noodle=300=[^[:space:]]*/--noodle=300=/g' ./test.con
f
+ [[ 0 -ne 0 ]]
```
This is the part of the code that is messing it up: `key=--noodle=300`, `sed` is looking for `noodle` not `--noodle`.

I think `sed` is also part of the problem, specifically this part `--noodle=300=[^[:space:]]*/--noodle=300=`, There is an extra `=`.  Why is it matching `--noodle=300=`?

atar-axis:
Hum, I am a bit confused!

> This is the part of the code that is messing it up: key=--noodle=300, sed is looking for noodle not --noodle

sed **should** only look for `noodle`! That's correct!

In shell scripts there is a convention that long parameters/argument-names are passed via `--` to the script, short ones are passed via `-`, for example:

`commandname --help` vs. `commandname -h`

In your script you have to remove those leading dashes because modprobe does expect the values to be in the format `key=value`

CodeCanna:
>sed should only look for noodle! That's correct!

However the value is `--noodle` not `noodle`.  So I have to use something like `tr` to remove those dashes, and turn the other dashes into unserscores?

ex:
`--debug-level=0` `tr` to `debug_level=0`

atar-axis:
I wouldn't automate that but do the mapping manually. When --noodle-doodle is called you could also change the value `Spaghetti` instead!

CodeCanna:
Could you go more in depth with what you mean?  I'm sorry, I don't understand what you are saying.  :confused:

And I also want to say, thanks for being patient with me, I know that I ask LOTS of questions.  That's just because I want to get this working the best I can, and I want to REALLY learn these concepts.  I have learned more in the past few weeks than in 2 years of tutorials, thanks for that!...Ok now that my sentimental crap is out of the way.

Do you mean something like this:
`./configure.sh --noodle-doodle=3`
```
if [[  "$key" == "--noodle-doodle" ]];
then
    key=noodle_doodle
fi
```

atar-axis:
>Do you mean something like this:
>./configure.sh --noodle-doodle=3
>
>```
>if [[  "$key" == "--noodle-doodle" ]];
>then
>    key=noodle_doodle
>fi
>```

Looks right! In my eyes you should check explicitly for parameters like this:

```
#!/bin/bash

cook()
{
  TYPE=$1
  SUBTYPE=$2
  shift; shift;
  echo "i am going to cook $TYPE today, strictly speaking $SUBTYPE"
}

for i in "$@"
do
case $i in
    -n=*|--noodles=*)
    KEY=pasta # !!! This is the key, it is hardcoded !!!
    VALUE="${i#*=}" # do validate the value before passing it!
    cook $KEY $VALUE
    shift # past argument=value
    ;;
    *)
          # unknown option
    ;;
esac
done
```

**Warning!** this script is untested since I am currently on my Windows machine, but I think it should be clear what I mean :) Passing `--noodle=...` to the script should not mean that the changed value in the config file is limited to `noodle=...`. Instead, you could also change `pasta=...`! You therefore somehow **translate** the parameters to the config-keys.

CodeCanna:
Ahhhhhh I think I see what you are doing here.  Can you elaborate more on this `VALUE="${i#*=}`?  

`i#` is the number of arguments right?  And what's the `*=`?

CodeCanna:
```
for i in "$@"
do
case $i in
    -n=*|--noodles=*)
    KEY=pasta # !!! This is the key, it is hardcoded !!!
    VALUE="${i#*=}" # do validate the value before passing it!
    cook $KEY $VALUE
    shift # past argument=value
    ;;
    *)
          # unknown option
    ;;
```
This is how I did the args before, in my first version of this script.  Then I used getopt, should I not use getopt?

atar-axis:
It is called **parameter expansion**, quite handy but very cryptic.
You can easily observe what it does here:

`PARAM="--bla=blub" echo ${PARAM#*=};`
returns: `blub`

It is like regexp, but it has an own syntax:
PARAM is the text on which the following will be done:
* `#` make the match as short as possible
* `*` delete the match
* pattern to look for: `=`
So it will read out everything from the variable PARAM. It does then search for the *first* `=` and deletes anything up to there. The result is <strike>--bla=</strike>blub.

For more information see here:
https://wiki.bash-hackers.org/syntax/pe
https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

atar-axis:
> This is how I did the args before, in my first version of this script. Then I used getopt, should I not use getopt?

You can use getopts if you want, I don't really care :smile: - up to you!

atar-axis:
What I wanted to emphasize is that it doesnt Matter what the argument passed to your script is called, you can even use `configure.sh --noodle=spaghetti` to change the `disable_ff` value (Not the noodle value). 

Hope that's clear now :)

CodeCanna:
Ok so I think I'm getting somewhere, check this out.  The only problem is I get an error `./configure.sh: line 339: ${$OPTS#*=}: bad substitution`, but I feel like this is on the right track.

Code:
```
-d | --debug-level)
        key=debug_level
        value="${$OPTS#*=}"
        debug_level $@ $key $value  #set_option() is called in this function.
        shift 2
        ;;
```

atar-axis:
Try {OPTS#*=} instead

atar-axis:
btw, you are free to use getopts! Use whatever works for you.
If you want to use something without `getopts`: Use this as a starting point
```
#!/bin/bash

change_disable_ff_param()
{
    echo "changing the parameter disable_ff, new value: $1"
}

for i in "$@"
do
case $i in
    -d=*|--disable_ff=*)
    VALUE="${i#*=}"
    set_disable_ff_param $VALUE 
    shift
    ;;
    *)
          # unknown option
    ;;
esac
done

if [[ -n $1 ]]; then
    echo "Last line of file specified as non-opt/last argument:"
    tail -1 $1
fi
```

CodeCanna:
Ok so now I'm getting an error because when i get `$value`, for example:

`--debug-level=3`, stores ```'3' --``` in `$value`, so now `debug_level()` doesn't work.  Is there a way we can add to `VALUE="${i#*=}` to get rid of the `--`?  I have tried this:

```
key="debug_level${OPTS#*=}"
        #value="${OPTS#*=}"
        sub_value=$(echo $value | tr -d "-")
        #debug_level $@ $key $value  #set_option() is called in this function.
        echo "$key"
        shift 2
        ;;
```

But this only gets me `debug_level --debug-level '3' --`, and this `$value=$(echo $value | tr -d "-")` gets me `debuglevel '3' `

atar-axis:
Sorry I am not sure what you are trying to do, can you give me a runable minimalistic version of your script?

CodeCanna:
So funny thing, while I was writing the example script that I was going to post here.  I figured it out!  Check this out, I was using `$OPTS` in `value="${OPTS#*=}"`, I should have used ```$2``` instead, which I did above but I didn't know about not needing the `$` directly in front of the variable name.  It looks like its working!  YESSSSSSSS!

Now to just do this for the rest of the options and it should be good to go!

atar-axis:
Sounds good :)

CodeCanna:
Hey for `combined_z_axis` does the value of the option have to be `y` or `n`, `0` or `1`?

atar-axis:
it doesn't matter!

CodeCanna:
Ok, check it out and see what you think.

CodeCanna:
Ok, now it seems it wants me to use brace expansion for the `$DETECTE_VERSION`.  Then that looks like its it.

atar-axis:
hey there, I will give it a try a bit later - for now: is there a reason why you are using a new way to detect the installed Versions? We already have one in `update.sh`:

`INSTALLED=$(dkms status 2>/dev/null | grep '^hid-xpadneo' 2>/dev/null | sed -E 's/^hid-xpadneo, ([0-9]+.[0-9]+.[0-9]+).*installed/\1/')`

atar-axis:
Yeah, just give the suggestion a try - it's not super important to do it this way but it looks a bit clearer.
I will give it a try a bit later

CodeCanna:
No really a reason, that's just the way I grabbed the version number from the file name.  I will use the one in `update.sh`.  It doesn't seem to be working for me though.

`DETECTED_VERSION=$(dkms status 2>/dev/null | grep '^hid-xpadneo' 2>/dev/null | sed -E 's/^hid-xpadneo, ([0-9]+.[0-9]+.[0-9]+).*installed/\1/')`

when i echo `$DETECTED_VERSION`, its empty.

atar-axis:
that's not so important, keep your way - we can do that later ;)

atar-axis:
Hey hey, I will give it a try today :) Thank you for your work!

atar-axis:
I like it! It works like a charm so far :)
Haven't had a look into the code but I will do that later.

One thing I would like to mention:
`combined_z_axis=0` is the default value which you are echoing, but `combined_z_axis=y/n` is the one which you are replacing. keep it consistent please :)

CodeCanna:
>I like it! It works like a charm so far :)

That is so great to hear!!

>Thank you for your contribution so far!

It was my pleasure and my joy, I have learned SO much in the past month or so working on this.  It's been great!

>Are you going to fix the Shellcheck issues? Tell me if there is a reason to prefer the existing way over the one mentioned by codacy/code climate

There is no preference, that's just the only way I know to do it, but I would love to learn a better way.  I tried copying this line from `update.sh`:

`VERSIONS=($(dkms status 2>/dev/null | grep '^hid-xpadneo,' 2>/dev/null | sed -E 's/^hid-xpadneo, ([0-9]+.[0-9]+.[0-9]+).*/\1/'))`

however I could not get it to work right.  When I echo `$VERSIONS` its empty, I am probably not doing it right.  I'll mess around with it again and show you what I am trying.

>combined_z_axis=0 is the default value which you are echoing, but combined_z_axis=y/n is the one which you are replacing. both are fine, but just keep it consistent please :)

I can fix that no problem.

>Ah and another thing: If the user does run the script without any option then - at least in my eyes - the help message should be shown too, right? So ./configure.sh -h should behave the same as ./configure.sh

This is a great idea, I will add this as well.  I never thought of it but it is true that most commands show the help menu when no arguments are called.

CodeCanna:
So I took that line from the `update.sh` and put it in my script and echoed it and it is empty.

atar-axis:
What's the Output of `dkms status` in your system

CodeCanna:
I'm out and about, I'll message you when I get home.

CodeCanna:
Output of `dkms status`:
```
Error! Could not locate dkms.conf file.
File: /var/lib/dkms/hid-xpadneo/0.5.4/source/dkms.conf does not exist.
```

Looks like I'm missing something.  I'll see what I can find as far as a fix.

CodeCanna:
A few possible fixes are [this](https://bbs.archlinux.org/viewtopic.php?id=237896) and [this](https://bbs.archlinux.org/viewtopic.php?id=151965).  Unfortunately I don't know much about dkms yet, but I'll read through these to see if I can find anything.

atar-axis:
Have you tried to run the Install Script before?

atar-axis:
1. Run the Uninstall Script
2. Remove /var/lib/dkms/xpadneo...
3. Run dkms Status again

CodeCanna:
Ok I ran the uninstall script and it said it found 0 instances of xpadneo.
I removed `/var/lib/dkms/hid-xpadneo/`
I ran dkms status and now it outputs nothing.

atar-axis:
Perfect, run the Install Script now - it should Work then

CodeCanna:
Nice!  It works now.  I will replace that line and push that.

atar-axis:
Perfect :) I will merge your work into master tomorrow/later - I think it is ready now :) See ya later!

CodeCanna:
Sounds great to me, thanks for you patience and letting me work on this.  If I have an idea for your driver do I open a new issue?  You might already be working on what I am thinking though.  I was thinking that it would be great if this worked with the wireless USB dongle instead of just bluetooth.  Let me know what you think.

atar-axis:
Merged by 769429bc91ca9883474008cc330721f2ef8ea638

PR #98, opened 2019-04-28
New udev rule

atar-axis:
Tested on:
* Kernel 5.0.x


atar-axis:
I will merge this PR in 7h if no objections are raised.

atar-axis:
**For Documentation:**
If this UDEV rule still makes problems, then we should maybe switch to one which fires a bit later (when the `event` file is created).

PR #103, opened 2019-05-09
added pydev example, closes #102

### LINKED ISSUE #102: Provide example script in Python to start a rumble effect on the xbox controller
**Is your feature request related to a problem? Please describe.**
I don't get a rumble effect when trying to make the xbox controller rumble using the example python script provided on the evdev documentation site:
[https://python-evdev.readthedocs.io/en/latest/tutorial.html#injecting-an-ff-event-into-first-ff-capable-device-found](url)

I am testing on a raspberry py raspbian stretch, kernel 4.14.98-v7+, python3.6 and the xpadneo driver installed. The controller is wireless using bluetooth. Rumble should in principle be possible since the tool fftest works fine and the controller rumbles, when xpadneo takes over control :-)

**Describe the solution you'd like**
A python example script using **evdev** would be perfect, since i am using evdev together with **asyncio** to read out the controller buttons and analog input.

**Describe alternatives you've considered**
An alternative would be a solution, where one uses e.g. **fcntl.ioctl** to write directly to the /dev/input/eventx file.
An example for that is given in 
[https://github.com/chrippa/ds4drv/issues/91](url)
In my case i get en error and i am currently unable to understand/remove the error.

atar-axis:


atar-axis:
close #102

PR #104, opened 2019-05-15
Create gamepad.py

gordonsolar:
example class to use the xbox controller with python asyncio (asynchronous mode, coroutines) and evdev
not all buttons and triggers are included in this class - it is an example only.
python>=3.6 is necessary
script is tested on a raspberry pi 3


atar-axis:
Hey there!
Thank you for the PR! Could you please fix the Problems (trailing whitespaces) before I merge it :)

atar-axis:
Nearly :)

Issue #99, opened 2019-05-05
Use Gamepad as Mouse on GUI


ghost:
**Summary:**
I'd love having the ability to control Linux GUI using the XBox One Controller and xpadneo driver as a mouse substitute.

**Specifics:**
I'm currently running Raspbian OS on a Raspberry Pi 3 B+. 
Using Parsec i created a streaming box to stream games from my computer to my TV via LAN.
My XBox One Controller is hooked up to the Raspberry Pi via bluetooth and works like a charm in game with xpadneo.

Now i'd love to use the Gamepad as a mouse on Raspbian so i don't need a physical mouse hooked up to the Raspberry Pi anymore to control the Raspbian GUI.

**Here's a suggestion for a possible key assignment:**

Left Joystick = cursor movement
right joystick = scroll wheel
button "A" = left click
button "B" = enter 
button "X" = right click
"menu" button = maybe a shortcut to the linux shutdown/restart window?

(Scrolling could also be done using the left and right shoulder buttons.)

atar-axis:
Hey there Lukas ;)
I would suggest to use the menu Button to Switch between the Mouse and gamepad behaviour. 

There are already some Projects around (QJoyPad and Antimicro e.g.). I am not sure therefore If it is a good Idea to so this in kernelspace.  I will give it a try and think about it - stay tuned :)

atar-axis:
Okay, it is quite easy to do it using antimicro :)

![Peek 2019-05-05 20-39](https://user-images.githubusercontent.com/11148588/57198689-f3951000-6f75-11e9-951d-ec05eff1f31d.gif)

atar-axis:
It is quite easy to do this using antimicro :)
I wasn't able to record my screen for some reason, but I think the following screenshot may explain it too:
![image](https://user-images.githubusercontent.com/11148588/57198738-abc2b880-6f76-11e9-98e0-d8eb95578c7d.png)


1. Connect your gamepad
2. Launch AntiMicro
3. Click on the `L-Stick Up` Button
4. Select the `Mouse` Tab
5. Press the big `Up` button
6. Do the same for the other directions and buttons/keys

FreakyAlleyCat:
That looks awesome, will give it a shot and report back! :)

atar-axis:
I thought a bit about it and I think I will not implement this in this driver, the reason is redundancy in essence. I would bloat up the kernel if every driver translates the gamepad events into mouse events itself.
[...]

PR #105, opened 2019-05-16
Virtual Mouse: Use the gamepad as mouse

atar-axis:
As described in #99


atar-axis:
**TODO**
* deactivate vmouse communication in gamepad mode
  * register
  * unregister
  * suspend
* implement deadband (for gp and mouse mode)
* only use vmouse module when a given parameter is set, so users do not need the vmouse module in general

PR #131, opened 2019-09-01
Update README.md

HaasJona:
Fedora additionally needs `make` for installation (which isn't a dependency of the other packages)
Otherwise you get

```
Building module:
cleaning build area...(bad exit status: 127)
make -j12 KERNELRELEASE=5.2.9-200.fc30.x86_64 -C /lib/modules/5.2.9-200.fc30.x86_64/build M=/var/lib/dkms/hid-xpadneo/0.6.0/build/src modules...(bad exit status: 127)
Error! Bad return status for module build on kernel: 5.2.9-200.fc30.x86_64 (x86_64)
Consult /
```

and the logfile says `make` is missing.


atar-axis:
Thanks!

PR #136, opened 2019-09-29
Correct disable_ff values

ipkpjersi:
There can only be 0 (enabled) and 1 (disabled) for disable_ff.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -51,9 +51,9 @@ function check_param {
         fi
         ;;
     "disable_ff")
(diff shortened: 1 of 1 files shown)


atar-axis:
Sorry, that's wrong. Any value between 0 and 3 is valid since the number is interpreted as a bitmask:
#define PARAM_DISABLE_FF_NONE    0
#define PARAM_DISABLE_FF_MAIN    1
#define PARAM_DISABLE_FF_TRIGGER 2
#define PARAM_DISABLE_FF_ALL     3

PR #138, opened 2019-10-03
Fix build and debug print for kernel version >= 5.4

PiMaker:
Build was broken in 5.4. This should fix it.

I tested this on mainline 5.4-rc1 and it worked fine, haven't tested on anything else, but since the hid_dbg macro was not affected by the commit mentioned, it should work fine for older kernels as well.


atar-axis:
Thanks! Looks like they removed `hid_printk`, yap.

PR #178, opened 2020-05-03
Several experimental improvement patches

kakra:
This series of commits is a work-in-progress code and quality improvement.

Currently it has some patches that are meant to improve connection stability and adhere to the controller specs better.

Could the reporters of the issues linked in these commits please comment on how it affects their local installations?

For me, the controller seems to be mostly stable now in gaming sessions of 10 hours or more. The only remaining issue is a disconnect during the first minutes of gameplay:

As far as I could observe, the controller no longer disconnects immediately with the start of the rumble but usually the rumble now plays and fails to stop (it continues to play unmodified no matter what the game wants to do with the rumble) while I'm still able to control the gameplay. But then, usually after a few seconds or a minute, it eventually freezes completely and then disconnects after a timeout or when I pull batteries. After reconnect, the issue does not occur again. This suggests that there is a quirk somewhere in the firmware that we should work around.

The upstream kernel in `hid-microsoft.c` uses a dedicated kernel work-queue for playing the effects so this may be an option to implement here. But the code needs some serious clean-up which is work-in-progress in this branch.

**Update:** @atar-axis I've done a major overhaul of the code base, reduced complexity a lot and adjusted the driver to better follow the design of other kernel drivers.

 **Update:** Work-queues for battery and rumble reports are implemented.


atar-axis:
Feel free to merge it, the only thing I would change is excluding the `.editorconfig` and `.indent.pro` - but I don't care ;)

kakra:
> Feel free to merge it, the only thing I would change is excluding the `.editorconfig` and `.indent.pro` - but I don't care ;)

I added these to make authors comply with kernel coding standards automatically. This could ideally be made into a Git pre-commit hook. But I didn't feel THAT adventurous.

PR #193, opened 2020-05-27
Always use curly braces and breaks out of switch cases.

adam-becker:
While reading through the code I noticed that the formatting didn't seem entirely consistent. Not sure if patches like this are accepted, but I threw this together anyways.

### DIFF
--- a/hid-xpadneo/src/hid-xpadneo.c
+++ b/hid-xpadneo/src/hid-xpadneo.c
@@ -402,9 +402,9 @@ static int xpadneo_get_battery_property(struct power_supply *psy,
 		    || xdata->psy_desc.type == POWER_SUPPLY_TYPE_UNKNOWN)
 			val->intval = POWER_SUPPLY_CAPACITY_LEVEL_UNKNOWN;
 		else if (XPADNEO_BATTERY_MODE(flags)
-			 || XPADNEO_BATTERY_CHARGING(flags)) {
+			 || XPADNEO_BATTERY_CHARGING(flags))
 			val->intval = capacity_level_map[level];
-		} else
+		else
 			val->intval = POWER_SUPPLY_CAPACITY_LEVEL_UNKNOWN;
 		break;
 
@@ -608,9 +608,10 @@ static void xpadneo_update_battery(struct xpadneo_devdata *xdata, u8 value)
 		break;
 	default:
 		xdata->psy_desc.type = POWER_SUPPLY_TYPE_UNKNOWN;
+		break;
 	}
 
-	if (XPADNEO_BATTERY_CHARGING(value) || XPADNEO_BATTERY_PRESENT(value))
+	if (XPADNEO_BATTERY_CHARGING(value) || XPADNEO_BATTERY_PRESENT(value)) {
 		switch (XPADNEO_BATTERY_CAPACITY_LEVEL(value)) {
 		case 0:
 			xdata->battery_capacity = 5;
@@ -625,6 +626,7 @@ static void xpadneo_update_battery(struct xpadneo_devdata *xdata, u8 value)
 			xdata->battery_capacity = 99;
 			break;
 		}
+	}
 
 	power_supply_changed(xdata->battery);
 }
@@ -708,7 +710,6 @@ static int xpadneo_input_configured(struct hid_device *hdev,
 			 (u16)xdata->idev->id.product);
 		xdata->idev->id.product = 0x02E0;
 		break;
-
 	}
 
 	if (param_combined_z_axis) {
--- a/hid-xpadneo/src/hid-xpadneo.c
+++ b/hid-xpadneo/src/hid-xpadneo.c
@@ -777,13 +777,13 @@ static int xpadneo_probe(struct hid_device *hdev,
 	ret = hid_parse(hdev);
 	if (ret) {
 		hid_err(hdev, "parse failed");
-		goto return_error;
+		return ret;
 	}
 
 	ret = hid_hw_start(hdev, HID_CONNECT_DEFAULT & ~HID_CONNECT_FF);
 	if (ret) {
 		hid_err(hdev, "hw start failed\n");
-		goto return_error;
+		return ret;
 	}
 
 	ret = xpadneo_init_ff(hdev);
@@ -791,9 +791,6 @@ static int xpadneo_probe(struct hid_device *hdev,
 		hid_err(hdev, "could not initialize ff, continuing anyway\n");
 
 	return 0;
-
-return_error:
-	return ret;
 }
 
 static void xpadneo_release_device_id(struct xpadneo_devdata *xdata)


kakra:
Actually, the code style tries to follow kernel style. And kernel style says that you don't use curly braces for single line if blocks:

https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces
> Do not unnecessarily use braces where a single statement will do.

The initial example in the same section actually shows an example with curly braces but I'm not sure how to interpret this: It's not clear if this is an example or a recommendation. In the end I used the code style script from the kernel and it removed those braces.

Similar of case statements: The kernel recommends to not use curly braces there.

So thanks for your suggestion, that's probably also how I would write my code but it doesn't match kernel style.

But if you want to work on this anyways: I'm not that happy with some of the complexity and encapsulation of the code. Usually, if you need more than 3 or 4 indents, that indicates code smell and too deep nesting. Some code blocks you amended are a good candidate for this. That should be split into different functions, maybe encapsulating the code logic differently. I'd happily look at such PRs.

adam-becker:
Ah, I think there are still some spots that don't follow those guidelines. Let me see if I can match that better.

On the issue of too much nesting: I might be willing to look into that as well, but it'll have to be prioritized with everything else in my backlog so it might be a while.

adam-becker:
Okay, I think this follows the guide. Not sure how the cases should be handled still, if the last case should always be drop-through.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -240,15 +240,14 @@ static int xpadneo_ff_play(struct input_dev *dev, void *data,
 	 * so we can completely turn off the main rumble while still doing
 	 * trigger rumble alone
 	 */
-	if (direction <= DIRECTION_UP) {
+	if (direction <= DIRECTION_UP)
 		/* scale the main rumbling between 0x0000..0x8000 (100%..0%) */
 		fraction_MAIN =
 		    ((DIRECTION_UP - direction) * 100) / DIRECTION_UP;
-	} else {
+	else
```
kakra, line 247: I don't think we should omit braces when a comment is involved. It enables to pitfalls too easily.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -402,9 +401,9 @@ static int xpadneo_get_battery_property(struct power_supply *psy,
 		    || xdata->psy_desc.type == POWER_SUPPLY_TYPE_UNKNOWN)
 			val->intval = POWER_SUPPLY_CAPACITY_LEVEL_UNKNOWN;
 		else if (XPADNEO_BATTERY_MODE(flags)
-			 || XPADNEO_BATTERY_CHARGING(flags)) {
+			 || XPADNEO_BATTERY_CHARGING(flags))
 			val->intval = capacity_level_map[level];
```
kakra, line 405: I'm not sure about this one... It looks like the kernel formatting script automatically added the braces here. Could you check by running `indent` on the source? There's an `.ident.pro` profile which is used when you run `indent` from the project root (aka `indent -v hid-xpadneo/src/hid-xpadneo.c`).

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -608,9 +607,10 @@ static void xpadneo_update_battery(struct xpadneo_devdata *xdata, u8 value)
 		break;
 	default:
 		xdata->psy_desc.type = POWER_SUPPLY_TYPE_UNKNOWN;
+		break;
 	}
 
-	if (XPADNEO_BATTERY_CHARGING(value) || XPADNEO_BATTERY_PRESENT(value))
+	if (XPADNEO_BATTERY_CHARGING(value) || XPADNEO_BATTERY_PRESENT(value)) {
```
kakra, line 613: Same here as above, I think the kernel formatting script added these again after I removed them. This probably also serves as a counter-measure against easy pitfalls. Also, technically speaking, the switch case statement isn't really a single statement. It's more like combined goto and if on steroids. I would argue differently if it wouldn't implicitly fall through.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -608,9 +607,10 @@ static void xpadneo_update_battery(struct xpadneo_devdata *xdata, u8 value)
 		break;
 	default:
 		xdata->psy_desc.type = POWER_SUPPLY_TYPE_UNKNOWN;
+		break;
```
kakra, line 610: I'm undecided about this. I don't think "default" should have an explicit break. You either put it last to have a fallback, or you put it first to have a common initializer.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -776,23 +776,20 @@ static int xpadneo_probe(struct hid_device *hdev,
...
-		goto return_error;
+		return ret;
 	}
 
 	ret = xpadneo_init_ff(hdev);
 	if (ret)
 		hid_err(hdev, "could not initialize ff, continuing anyway\n");
 
 	return 0;
-
```
kakra, line 793: Yeah, LGTM :-)

hid-xpadneo/src/hid-xpadneo.c
adam-becker: What pitfalls, precisely?

hid-xpadneo/src/hid-xpadneo.c
adam-becker: I believe fall through in any case is bad practice, and whatever is trying to be achieved with it can be achieved in other, simpler ways. It always comes as a surprise when I'm reading through code.

hid-xpadneo/src/hid-xpadneo.c
kakra: The indented part already splits across multiple lines. One may be tempted to actually split the statement within the indention without the compiler noticing any problems, and neither the human (since it looks fine indention-wise).

hid-xpadneo/src/hid-xpadneo.c
kakra: Okay, then let's accept this change.

hid-xpadneo/src/hid-xpadneo.c
kakra: But I agree that "} else" looks ugly. I'm pretty sure that `indent` did this.

adam-becker: Running indent just put spaces between the curly braces on lines 817, 818, 821 and 824

adam-becker:
```
$ indent --version
GNU indent 2.2.12
```

kakra: Okay, let's pick this change and I'll cross-check with my local indent version.

adam-becker:
Just noticed while browsing, but the first switch in `xpadneo_update_battery` has a `case 0x10` that can't actually be hit. The switched value is between 0 and 3, so either the case is wrong or the switched value is.

kakra:
> Just noticed while browsing, but the first switch in `xpadneo_update_battery` has a `case 0x10` that can't actually be hit. The switched value is between 0 and 3, so either the case is wrong or the switched value is.

Oh wow... Probably because the later the night the dumber the programmer. I was probably thinking binary. ;-) You earned a free ticket for fixing this in a separate commit.

hid-xpadneo/src/hid-xpadneo.c
adam-becker: So should a comment count towards the number of statements in an `if` then? I can put the braces back in here.

adam-becker: The braces here are back in.

kakra: > So should a comment count towards the number of statements in an `if` then? I can put the braces back in here.

Yes, I think that's better for authors not yet familiar with the code. Especially long comments may confuse people about the curly braces.

kakra:
Could you squash the formatting commits?

adam-becker:
Squashed the 2 formatting commits

hid-xpadneo/src/hid-xpadneo.c
kakra: I got some time to remember... Yes, I think ident was confused about that part and put spaces into strange places. Maybe that should be refactored. I'll probably look at your PR tomorrow. If you have any updates on this, feel free to rebase and push.

PR #195, opened 2020-05-29
Fix bug where 2 would map to unknown PS type instead of battery.

adam-becker:
Fixing a bug found in my first PR.


kakra:
Thanks.

PR #196, opened 2020-05-29
Add option for FF connect notification.

### LINKED ISSUE #149: Option to disable rumble connection notification
**Is your feature request related to a problem? Please describe.**
Please provide option do disable rumble effect that notifies you about successful connection to the device.

**Describe the solution you'd like**
An additional option in configuration menu to disable rumble connection notification (as opposed to disabling rumble altogether) 

**Describe alternatives you've considered**
NA

**Additional context**
NA

adam-becker:
Closes #149

### DIFF
--- a/hid-xpadneo/src/hid-xpadneo.c
+++ b/hid-xpadneo/src/hid-xpadneo.c
@@ -66,6 +66,11 @@ module_param_named(trigger_rumble_damping, param_trigger_rumble_damping, byte,
 MODULE_PARM_DESC(trigger_rumble_damping,
 		 "(u8) Damp the trigger: 1 (none) to 2^8+ (max).");
 
+static bool param_ff_connect_notify;
+
+module_param_named(ff_connect_notify, param_ff_connect_notify, bool, 0644);
+MODULE_PARM_DESC(ff_connect_notify,
+		 "(bool) Connection notification using force feedback. 1: enable, 0: disable.");
 
 static DEFINE_IDA(xpadneo_device_id_allocator);
 
@@ -361,7 +366,7 @@ static int xpadneo_init_ff(struct hid_device *hdev)
 	if (xdata->output_report_dmabuf == NULL)
 		return -ENOMEM;
 
-	if (!param_disable_ff)
+	if (!param_disable_ff && param_ff_connect_notify)
 		xpadneo_welcome_rumble(hdev);
 
 	input_set_capability(idev, EV_FF, FF_RUMBLE);
--- a/hid-xpadneo/src/hid-xpadneo.c
+++ b/hid-xpadneo/src/hid-xpadneo.c
@@ -66,7 +66,7 @@ module_param_named(trigger_rumble_damping, param_trigger_rumble_damping, byte,
 MODULE_PARM_DESC(trigger_rumble_damping,
 		 "(u8) Damp the trigger: 1 (none) to 2^8+ (max).");
 
-static bool param_ff_connect_notify;
+static bool param_ff_connect_notify = 1;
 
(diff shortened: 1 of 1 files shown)


hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -57,13 +57,20 @@ static bool param_combined_z_axis;
 
 module_param_named(combined_z_axis, param_combined_z_axis, bool, 0644);
 MODULE_PARM_DESC(combined_z_axis,
-		 "(bool) Combine the triggers to form a single axis. 1: combine, 0: do not combine");
+		 "(bool) Combine the triggers to form a single axis. 1: combine, 0: do not combine.");
 
 static u8 param_trigger_rumble_damping = 0;
+
```
kakra, line 63: This empty line does not match the style of the block just above.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -57,13 +57,20 @@ static bool param_combined_z_axis;
 
 module_param_named(combined_z_axis, param_combined_z_axis, bool, 0644);
 MODULE_PARM_DESC(combined_z_axis,
-		 "(bool) Combine the triggers to form a single axis. 1: combine, 0: do not combine");
+		 "(bool) Combine the triggers to form a single axis. 1: combine, 0: do not combine.");
```
kakra, line 60: This change is unrelated and should be a separate commit, you may put it in the same PR, tho. The same for the change a few lines below.

hid-xpadneo/src/hid-xpadneo.c
adam-becker: The style on this was already inconsistent before my PR, which one is preferred?

adam-becker:
Also, does this need to be enabled by default?

hid-xpadneo/src/hid-xpadneo.c
kakra: I think we should stick with how in-kernel drivers do it.

kakra:
> Also, does this need to be enabled by default?

The notification should play by default so we can easily catch clones not supporting the rumble protocol correctly.

hid-xpadneo/src/hid-xpadneo.c
adam-becker: I'm not sure what that looks like. The code in `master` doesn't follow a consistent style either.

kakra: I'll take a peek when I get to work on the code next time and fix this properly.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -57,13 +57,20 @@ static bool param_combined_z_axis;
...
 
 static u8 param_trigger_rumble_damping = 0;
+
 module_param_named(trigger_rumble_damping, param_trigger_rumble_damping, byte,
 		   0644);
 MODULE_PARM_DESC(trigger_rumble_damping,
-		 "(u8) Damp the trigger: 1 (none) to 2^8+ (max)");
+		 "(u8) Damp the trigger: 1 (none) to 2^8+ (max).");
```
kakra, line 67: I'm not sure if this is correct anyways. I never closely looked at the damping code. But when anything, the range for u8 is probably 0..255 and not 1..256 (or whatever `2^8+` should mean).

adam-becker: I'll leave that work for a future PR, just wanted the descriptions to look the same.

adam-becker: Also, it looks like it has an effective range of `1..255`

PR #197, opened 2020-05-30
Rename 'xpadneo_mapping' to match other names

adam-becker:
The diff makes the alignment look off, but I promise that it isn't :smile:

### DIFF
--- a/hid-xpadneo/src/hid-xpadneo.c
+++ b/hid-xpadneo/src/hid-xpadneo.c
@@ -523,9 +523,9 @@ err_free_name:
 	return ret;
 }
 
-static int xpadneo_mapping(struct hid_device *hdev, struct hid_input *hi,
-			   struct hid_field *field, struct hid_usage *usage,
-			   unsigned long **bit, int *max)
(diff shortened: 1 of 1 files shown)


hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -836,7 +836,7 @@ MODULE_DEVICE_TABLE(hid, xpadneo_devices);
 static struct hid_driver xpadneo_driver = {
 	.name = "xpadneo",
 	.id_table = xpadneo_devices,
```
adam-becker, line 838: I left this one as-is, since it didn't appear to even be attempting to match the field name.

kakra: For kernel drivers, you should probably set your tab size to 8. Then just follow what `indent` makes from it.

kakra: I just checked, looks code. I fixed a minor line length issue on the way. Will merge.

PR #200, opened 2020-06-08
hid_xpadneo, quirks: Add "8Bitdo SF30 pro"

friesnadrink:
Sorry if this isn't the right way to do this. It is probably all 8Bitdo controllers with an xbox mode that will need the same quirk entry. I don't have the exact string for any others, so I didn't include them.


kakra:
Can you confirm the name string for this particular model (without the `+`)?

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -152,6 +152,9 @@ static const struct quirk xpadneo_quirks[] = {
 	DEVICE_NAME_QUIRK("8BitDo SN30 Pro+",
 			  XPADNEO_QUIRK_NO_PULSE | XPADNEO_QUIRK_NO_TRIGGER_RUMBLE |
 			  XPADNEO_QUIRK_NO_MOTOR_MASK),
+	DEVICE_NAME_QUIRK("8Bitdo SF30 Pro",
```
kakra, line 155: Please sort the entries alphabetically. Thanks. :-)

kakra:
I guess it would make more sense to match the OUI of that manufacturer instead of trying to match each model name... I'll push another PR.

> Sorry if this isn't the right way to do this. It is probably all 8Bitdo controllers with an xbox mode that will need the same quirk entry. I don't have the exact string for any others, so I didn't include them.

No sorry please, that's exactly the right way to do it, tho I'll probably obsolete your commit with my next PR. ;-)

kakra:
@friesnadrink Could you try with my #201 instead? It should match any 8BitDo-made controller.

PR #201, opened 2020-06-08
hid_xpadneo, quirks: Migrate 8BitDo quirks to OUI match

kakra:
This should match any 8BitDo controller.

Closes: https://github.com/atar-axis/xpadneo/pull/200

### DIFF
--- a/hid-xpadneo/src/hid-xpadneo.c
+++ b/hid-xpadneo/src/hid-xpadneo.c
@@ -141,17 +141,20 @@ struct xpadneo_devdata {
 
 struct quirk {
 	char *name_match;
+	char *oui_match;
 	u16 name_len;
 	u16 flags;
 };
 
 #define DEVICE_NAME_QUIRK(n, f) \
 	{ .name_match = (n), .name_len = sizeof(n) - 1, .flags = (f) }
+#define DEVICE_OUI_QUIRK(o, f) \
+	{ .oui_match = (o), .flags = (f) }
 
 static const struct quirk xpadneo_quirks[] = {
-	DEVICE_NAME_QUIRK("8BitDo SN30 Pro+",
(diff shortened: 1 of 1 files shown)


hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -141,17 +141,20 @@ struct xpadneo_devdata {
...
 	u16 flags;
 };
 
 #define DEVICE_NAME_QUIRK(n, f) \
 	{ .name_match = (n), .name_len = sizeof(n) - 1, .flags = (f) }
+#define DEVICE_OUI_QUIRK(o, f) \
+	{ .oui_match = (o), .flags = (f) }
 
 static const struct quirk xpadneo_quirks[] = {
-	DEVICE_NAME_QUIRK("8BitDo SN30 Pro+",
```
adam-becker, line 152: Now that this macro isn't used, it can be removed. Add it back in if necessary later.

kakra: It could but I didn't like the idea of digging it out later... The pure existence of the macro should make it much easier for contributors to add their devices via PR. So 2 points for keeping it vs. 1 to remove it. But I was also thinking about it.

kakra: Also, this PR isn't tested yet by owners of 8BitDo controllers. So maybe we end up with merging both PRs. All matching device flags will be ORed together so in theory we could refactor the flags to have common ones in the OUI match and model specific flags in the name match.

kakra:
Ping @friesnadrink

Could you test this? Do you own different 8BitDo controllers and test with both?

kakra:
Merged without feedback from original reporter, the change should work.

PR #204, opened 2020-06-20
Rumble protocol throttling and concurrency fixups

kakra:
Please test.


kakra:
Tested for multiple hours without problems.

PR #206, opened 2020-06-26
dkms: Eliminate bash scripting from udev rules

kakra:
@currybullen Please test on Fedora.

This commit fixes depending on calling modprobe from udev. Instead, we now install a modalias database to override the kernel aliases and use attribute writes to rebind the driver.

Closes: https://github.com/atar-axis/xpadneo/pull/167


hid-xpadneo/src/etc-udev-rules.d/98-xpadneo.rules
```diff
@@ -0,0 +1,2 @@
+ACTION=="add", KERNEL=="0005:045E:02FD.*|0005:045E:02E0.*|0005:045E:0B05.*", SUBSYSTEM=="hid", DRIVER!="xpadneo", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
+ACTION=="add", KERNEL=="js*", DRIVERS=="xpadneo", RUN+="/usr/bin/jscal -u 8,0,1,3,4,2,5,16,17,10,304,305,307,308,310,311,314,315,317,318"
```
kakra, line 2: This is not compatible with the Chrome Gamepad API as it tries to apply the mapping from `evdev` to `joydev`. Maybe better move to documentation instead.

kakra: Moved to patch queue for consideration.

kakra:
Let's look at no feedback as good feedback... Merged.

PR #213, opened 2020-07-01
Add support for XBE2 (Unknown mode)

ehats:
This commit includes the device descriptor, BT packet structure, as well
as the necessary fixups to get the axes and buttons to play along nicely.

## TODO

- [x] Remove unnecessary axes
- [x] Remove duplicate buttons
- [x] Fix button mapping


docs/descriptors/xbe2_unknown.md
```diff
@@ -0,0 +1,690 @@
...
+0x09, 0x31,        //     Usage (Y)
+0x15, 0x00,        //     Logical Minimum (0)
+0x27, 0xFF, 0xFF, 0x00, 0x00,  //     Logical Maximum (65534)
+0x95, 0x02,        //     Report Count (2)
+0x75, 0x10,        //     Report Size (16)
+0x81, 0x02,        //     Input (Data,Var,Abs,No Wrap,Linear,Preferred State,No Null Position)
+0xC0,              //   End Collection
+0x09, 0x01,        //   Usage (Pointer)
+0xA1, 0x00,        //   Collection (Physical)
+0x09, 0x32,        //     Usage (Z)  // FIXED
```
ehats, line 131: This is the original device descriptor, not the patched one. I'll remove my comments at the end, so please ignore these for now

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -217,6 +217,36 @@ static const struct usage_map xpadneo_usage_maps[] = {
 
 	/* disable duplicate button */
 	USAGE_IGN(0xC0224),
+
+	/* XBE2 hacks */
+	/* Disable "dial", which is a redundant representation of the D-Pad */
+	USAGE_IGN(0x10037),
+
+	/* Disable unnecessary axes Vx, Vy, Vz, Vbrx, Vbry, Vbrz which cause an additional axis to show up */
```
ehats, line 225: I thought it'd be semantically weird to only remove one of these, so I decided to just ignore all of them, but this _might_ actually be a bad idea - let me know what you think

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -743,12 +773,15 @@ static int xpadneo_input_mapping(struct hid_device *hdev, struct hid_input *hi,
 		}
 	}
 
+	hid_info(hdev, "MAP_AUTO: %lx\n", usage->hid);
```
ehats, line 776: Will be removed

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -743,12 +773,15 @@ static int xpadneo_input_mapping(struct hid_device *hdev, struct hid_input *hi,
...
+
 	/* let HID handle this */
 	return MAP_AUTO;
 }
 
 static u8 *xpadneo_report_fixup(struct hid_device *hdev, u8 *rdesc, unsigned int *rsize)
 {
+	hid_notice(hdev, "Descriptor size: %i\n", *rsize);
```
ehats, line 784: Will be removed

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -779,19 +812,19 @@ static u8 *xpadneo_report_fixup(struct hid_device *hdev, u8 *rdesc, unsigned int
...
-		    rdesc[144] == 0x29 && rdesc[145] == 0x0F &&
-		    rdesc[152] == 0x95 && rdesc[153] == 0x0F &&
-		    rdesc[162] == 0x95 && rdesc[163] == 0x01) {
-			hid_notice(hdev, "fixing up button mapping\n");
-			rdesc[145] = 0x0B;	/* 15 buttons -> 11 buttons */
-			rdesc[153] = 0x0B;	/* 15 bits -> 11 bits buttons */
-			rdesc[163] = 0x05;	/* 1 bit -> 5 bits constants */
-		}
-	}
+	// /* fixup reported button count for Xbox controllers in Linux mode */
```
ehats, line 815: Will add back, just wanted to start from a clean slate

ehats:
@kakra I think this is ready for some initial feedback. Status so far:
- Removed all unnecessary axes (2)
- Remove all of the weird button duplicates

hid-xpadneo/src/hid-xpadneo.c
kakra: Are one of the paddles analog axes? Or are they just digital?

hid-xpadneo/src/hid-xpadneo.c
kakra: It may be easier to add `#if 0` in front and `#endif` at the end of the block. Then you can turn it on and off by just swapping 0 with 1.

hid-xpadneo/src/hid-xpadneo.c
ehats: This was about the last axis, lagging behind LX.

All of the paddles are buttons, which appear in `evtest` as `type 1 (EV_KEY), code 240 (KEY_UNKNOWN), value 0` (where the value represent which paddles are pressed)

hid-xpadneo/src/hid-xpadneo.c
ehats: I just use `Ctrl+/` and VS code takes care of the (un)commenting :)

ehats:
Cool, I'll wait for your "so far so good" before getting started on remapping the buttons!

One thing that worries me though with all the ignoring - is it possible for me to ignore something that is actually useful for another device? :thinking:

hid-xpadneo/src/hid-xpadneo.c
kakra: Okay, they should become `BTN_*` symbols then. I think it's okay to remove the VR axes then.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -217,6 +217,36 @@ static const struct usage_map xpadneo_usage_maps[] = {
...
+	USAGE_IGN(0x10041),
+	USAGE_IGN(0x10042),
+	USAGE_IGN(0x10043),
+	USAGE_IGN(0x10044),
+	USAGE_IGN(0x10045),
+
+	/* Button 319 (copy of A button) */
+	USAGE_IGN(0x90010),
+	
+	/* TRIGGER_HAPPY1 (copy of B button) */
```
kakra, line 236: It doesn't make much sense to put the original symbol names here as these don't have a real meaning when reading the events - we are remapping stuff anyways. So it's sufficient to say "remove duplicate button A" in the final patch.

Any maybe also make this appear as one semantic block, put a header comment "remove duplicate buttons" and put one comment inline with each line "duplicate of A" etc (similar to the beginning of the array).

hid-xpadneo/src/hid-xpadneo.c
kakra: But the diff is bigger and sometimes harder to read... ;-) But it's fine - do what works for you.

hid-xpadneo/src/hid-xpadneo.c
kakra: Make the header comment shorter, like "disable blind axes" or something like that, then but the symbol name as a trailing comment on each line.

kakra:
> One thing that worries me though with all the ignoring - is it possible for me to ignore something that is actually useful for another device?

At some point we may need to pin usage maps to device models. Currently I don't see a value for that.

From the structure of the code, I think this is very easy to discover: It's right there in one comprehensible block, no need to dig through endless amounts of if-blocks. So if some new controller model or firmware struggles with it later, it's a matter of commenting it out to find the problem, and then we can apply some model-specific logic. But we don't know yet of what kind this requirement would be - so let's just leave it as it is for now.

ehats:
I believe I addressed all your comments regarding style. As for js/evtest output, what exactly would you like to see? One "event" for all of the buttons / axes for both jstest and evtest?

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -217,6 +217,26 @@ static const struct usage_map xpadneo_usage_maps[] = {
 
 	/* disable duplicate button */
 	USAGE_IGN(0xC0224),
+
+	/* XBE2 hacks */
```
kakra, line 221: Maybe put "XBE2" in front of each header, so we have that associated with each individual block instead of a semi-top-level header. It makes it easier to reorder blocks later. We should probably also add "Generic:" or "All:" to other block header comments.

It also makes it easier to follow up with some model specific logic later.

kakra:
> I believe I addressed all your comments regarding style. As for js/evtest output, what exactly would you like to see? One "event" for all of the buttons / axes for both jstest and evtest?

I think you can test all the buttons yourself for now... ;-) So I'd just would like to see the top of the output, like this:
```console
# jstest /dev/input/js0
Driver version is 2.1.0.
Joystick (Xbox Wireless Controller) has 8 axes (X, Y, Z, Rx, Ry, Rz, Hat0X, Hat0Y)
and 10 buttons (BtnA, BtnB, BtnX, BtnY, BtnTL, BtnTR, BtnSelect, BtnStart, BtnThumbL, BtnThumbR).
Testing ... (interrupt to exit)
Axes:  0:     0  1:     0  2:-32767  3:     0  4:     0  5:-32767  6:     0  7:     0 Buttons:  0:off  1:off  2:off  3:off  4:off  5:off  6:off  7:off  8:off  9:off                                                                       
```

ehats:
Alrighty, here's `jstest`:

```
Joystick (Xbox Elite Wireless Controller) has 8 axes (X, Y, Z, Rx, Ry, Rz, Hat0X, Hat0Y)
and 21 buttons (BtnA, BtnB, BtnX, BtnY, BtnTL, BtnTR, BtnSelect, BtnStart, BtnMode, BtnThumbL, BtnThumbR, (null), (null), (null), (null), (null), (null), (null), (null), (null), (null)).
```
And evtest is a bit more verbose:
[evtest.txt](https://github.com/atar-axis/xpadneo/files/4860421/evtest.txt)

Note: I only removed the buttons that were duplicates, I didn't touch the things that don't seem to be "pressable". Not really sure if we care about them, but I guess they're not the highest priority anyway

kakra:
> Alrighty, here's `jstest`:
> 
> ```
> Joystick (Xbox Elite Wireless Controller) has 8 axes (X, Y, Z, Rx, Ry, Rz, Hat0X, Hat0Y)
> and 21 buttons (BtnA, BtnB, BtnX, BtnY, BtnTL, BtnTR, BtnSelect, BtnStart, BtnMode, BtnThumbL, BtnThumbR, (null), (null), (null), (null), (null), (null), (null), (null), (null), (null)).
> ```

I think `BTN_MODE` must become `KEY_HOMEPAGE` otherwise games swap around `BtnMode`, `BtnThumbL`, and `BtnThumbR`. I don't yet know why, and at least Proton games do that, and there's no way of fixing it. I think that's because in Windows, this button is not exposed as a game-controller button, so games expect to see `BtnThumbL` right after `BtnStart`. And `KEY_HOMEPAGE` works the same for Steam, so that's okay, nothing is lost.

> And evtest is a bit more verbose:
> [evtest.txt](https://github.com/atar-axis/xpadneo/files/4860421/evtest.txt)
> 
> Note: I only removed the buttons that were duplicates, I didn't touch the things that don't seem to be "pressable". Not really sure if we care about them, but I guess they're not the highest priority anyway

Yeah, that has a complete PC101 keyboard map... It would be better to not put that in the array but just check for HID page 7 in the mapping function and ignore the block completely. Maybe put a comment that it may be needed should we add programming capabilities in the future.

ehats:
>  ignore the block completely.

Would that be just a check for `>= 0x70000 && < 0x80000` or is there a magic thing somewhere that just gets me the page ID?

I'll remove that keyboard thingy, then get started on fixing the X and Y buttons.

About BtnHome, I'm actually not yet if that actually exists as a button on the controller. It might be just one of the many "fake" buttons

kakra:
> > ignore the block completely.
> 
> Would that be just a check for `>= 0x70000 && < 0x80000` or is there a magic thing somewhere that just gets me the page ID?

You can use `(usage->hid && HID_USAGE_PAGE) == HID_UP_KEYBOARD` (see `hid.h` in the kernel). This way, no one has to remember later what page 7 is.

> I'll remove that keyboard thingy, then get started on fixing the X and Y buttons.

My button mapping fixup code actually changes the bit counts for data and padding. That already get's rid of a lot of buttons and may make some of your button ignores in the usage map obsolete. You could do something similar. No fear: The driver will still see the padding bits as data in the raw report, so it can shift the bits over into the non-padding bits (as I do with this `BIT(2)<<5` sort-of-magic).

> About BtnHome, I'm actually not yet if that actually exists as a button on the controller. It might be just one of the many "fake" buttons

After re-enabling the button fix-up code, it should trigger on the Xbox logo button. That button 11 of the original Xbox controller and should not be exposed as `BTN_MODE`. I used my `USAGE_MAP` macro to remap that to `KEY_HOMEPAGE`. Actually, in the browser you can then navigate to your start page with this button. ;-)

kakra:
Before applying any `jscal -u` remappings, you should test the controller in https://html5gamepad.com/, too. I'm not sure if Firefox works but Chrome does. Chrome and Firefox may actually show different mappings. I tested Chrome only, as that has Stadia.

kakra:
Do you think we can get this into shape for the 0.8 release? I'm planning to do one final 0.7 bugfix release, then add a few more tiny feature updates, plus maybe this PR, and release 0.8.

After 0.8 I'd like to rework the rumble code because jiffies timing in a kernel worker can be quite jittery. I've taken a different approach in the xow project and that seems to work a lot better. With this change, I'll also make it switchable by using a quirks mode for it - and that may introduce conflicting commits with your PR.

ehats:
When are you planning on releasing 0.8? I couldn't work on this yesterday and today, but I'm planning on resuming work on it over the weekend.

The other question is: what do we consider to be the first stopping point here? If we want the axes and buttons to be reported in a roughly sane way, I don't think that's too far off, but if we also wanna make sure that everything appears nicely in the higher levels (e.g. jstest, games, etc.), that's probably a bit more work, I'd expect?

kakra:
Then let's first see what we can make from this over the course of the weekend. I may be releasing 0.8 end of next week or the week after.

ehats:
So I set out this morning to deal with the X and Y buttons, and found some interesting things. This is the usage-physical button mapping on my controller:

|9000... | Button|
|-----------|---------|
|1 | A|
|2|B|
|4|X|
|5|Y|
|7|LB|
|8|RB|

Why is this interesting? Because of the already existing mapping code:
```
	USAGE_MAP(0x90001, MAP_STATIC, EV_KEY, BTN_A),	/* A */
	USAGE_MAP(0x90002, MAP_STATIC, EV_KEY, BTN_B),	/* B */
	USAGE_MAP(0x90003, MAP_STATIC, EV_KEY, BTN_X),	/* X */
	USAGE_MAP(0x90004, MAP_STATIC, EV_KEY, BTN_Y),	/* Y */
```

This is the wrong mapping my controller :) Is there a way for us to determine which controller we're working on in that part of the code so we can apply the correct mapping?

kakra:
Actually it's not. You are looking at it from the wrong perspective: we patch the raw HID packets to move button 4 bit to 3 (and others). Thus the usage map has to reflect this. This is needed to make the HID packets compatible to games expecting the original Xbox controller format.

ehats:
So does that mean I should be trying to change the descriptor to make sure that whenever I press the X button, that will translate to `0x90003`?

kakra:
This is the problem: The descriptor does not actually assign button codes, it just defines a bitmap of buttons. The Linux kernel will match each bit sequentially to `BTN_*`, thus it assign `BTN_A`, `_B`, `_C`, `_X`, `_Y`, `_Z`... There's actually no button C on the controller, so it will have some blind bits in its bitmap. What happens now in some games, is that X becomes Y, because some games look at the button number, expect button 4 to be Y, but in the HID fields it's actually button 5.

So the following fixups are needed: Reduce the number of buttons in the descriptor to the actual amount of buttons. Then in the raw reports shift the button bits to exclude the empty buttons, also, some button bits need to be re-ordered.

Finally, Linux would still expect the button bits to mean A, B, C, X, Y, Z, etc... So we need to tell that C does not exist and remap code 3 and 4 to X and Y (and so on for the rest of the buttons). This is what the `USAGE_MAP` macros do. Actually, that mapping is mandatory to stay compatible with all games (it defines the order in which Windows games expect the button codes, Windows games do not know about a C to Z button, and they expect no gaps in the maps). You will have to patch the raw HID reports and fixup the button count in the descriptor.

The following code patches the original Linux mode XB1S button usage page:
```
	/* fixup reported button count for Xbox controllers in Linux mode */
	if (*rsize >= 164) {
		/* 11 buttons instead of 10: properly remap the Xbox button */
		if (rdesc[140] == 0x05 && rdesc[141] == 0x09 &&
		    rdesc[144] == 0x29 && rdesc[145] == 0x0F &&
		    rdesc[152] == 0x95 && rdesc[153] == 0x0F &&
		    rdesc[162] == 0x95 && rdesc[163] == 0x01) {
			hid_notice(hdev, "fixing up button mapping\n");
			rdesc[145] = 0x0B;	/* 15 buttons -> 11 buttons */
			rdesc[153] = 0x0B;	/* 15 bits -> 11 bits buttons */
			rdesc[163] = 0x05;	/* 1 bit -> 5 bits constants */
		}
	}
```

The controller actually has 10 buttons for Windows games but we patch it to 11 buttons anyways to account for the Xbox logo button bit. We'll map that to a key later with `USAGE_MAP()` so that it's no longer counted as joystick button, that finally ends up as 10 buttons as reported by the joystick interface. If we report the Xbox logo button as a button, it will always be ordered before the thumb stick buttons by Linux and the joystick API, and that confuses games. Even in Windows, the logo button is a key and not a joystick button. `constants` is just padding in the descriptor and does not account for buttons, it only fills up unused space to a byte boundary.

The following code fixes the raw reports as received by the controller:
```
	/* correct button mapping of Xbox controllers in Linux mode */
	if (report->id == 1 && (reportsize == 17 || reportsize == 39)) {
		u16 bits = 0;

		bits |= (data[14] & (BIT(0) | BIT(1))) >> 0;	/* A, B */
		bits |= (data[14] & (BIT(3) | BIT(4))) >> 1;	/* X, Y */
		bits |= (data[14] & (BIT(6) | BIT(7))) >> 2;	/* LB, RB */
		bits |= (data[16] & BIT(0)) << 6;	/* Back */
		bits |= (data[15] & BIT(3)) << 4;	/* Menu */
		bits |= (data[15] & BIT(5)) << 3;	/* LS */
		bits |= (data[15] & BIT(6)) << 3;	/* RS */
		bits |= (data[15] & BIT(4)) << 6;	/* Xbox */
		data[14] = (u8)((bits >> 0) & 0xFF);
		data[15] = (u8)((bits >> 8) & 0xFF);
		data[16] = 0;
	}
```

It creates an empty 2-byte long bitmap in `bits`, and then it remaps the bits from the report to this new bitmap by shifting X,Y from bit 3,4 to bit 2,3 (counting from 0). LB,RB are the two buttons above the triggers, LS,RS are the two thumb stick buttons. You'll see that we shift some pairs of buttons differently for each member: The order of these is mandatory for games compatibility. In the end, we put the new bitmap back into the report. The reports are send to us for every input done on the controller in real time, while the descriptor is only received once during device init.

So, you cannot actually change the descriptor to remap the buttons as there is no way to define a mapping there. It's just a bitmap which Linux applies a standard logic to for giving it meaning: It just sequentially assigns all of it's `BTN_*` codes to each bit. But this logic differs from Windows, which doesn't have `BTN_C` and some of the others in its protocol for Xbox controllers.

I hope this clears it up a bit for you.

kakra:
After you have done this, you should've reached the following milestone:

1. There are no more blind bits when looking at the events in `evtest`: Each button should count values up one by one when pressing the buttons in the following order (`EV_MSC` events): A, B, X, Y, LB, RB, Back, Menu, LS, RS, Xbox logo.
2. `evtest` should have correctly assigned the `BTN_*` symbols to each of these events.
3. `jstest` orders the buttons in the same order but excludes the Xbox logo button.
4. `jstest` reports no blind buttons.
5. You may already report the additional buttons to user space and did give them a sensible meaning/symbol.

If this is true, the next thing to fix is the HTML5 gamepad tester (this may need to patch the PID of the controller), and SDL (this may need patching the PID or version number of the input device info):
```c
	case 0x02E0:
		hid_info(hdev,
			 "pretending XB1S Linux firmware version "
			 "(changed version from 0x%08X to 0x00000903)\n", xdata->idev->id.version);
		xdata->idev->id.version = 0x00000903;
		break;
	case 0x02FD:
		hid_info(hdev,
			 "pretending XB1S Windows wireless mode "
			 "(changed PID from 0x%04X to 0x02E0)\n", (u16)xdata->idev->id.product);
		xdata->idev->id.product = 0x02E0;
		break;
```

kakra:
> The following code fixes the raw reports as received by the controller:

BTW: The `report_size == 17` may be wrong here... I wasn't yet able to force the controller into this state to verify this. But a report with size 17 is the Windows mode report which may actually need to bit-shifting. Your reports are way bigger, but the bits for these buttons should be at the same position (I think the controller is backwards-compatible). Also, your controller may use other report IDs for its additional functions.

ehats:
I just pushed a new commit with some more changes that actually turn it into a very sane controller based on some very limited testing. 

> There are no more blind bits when looking at the events in evtest: Each button should count values up one by one when pressing the buttons in the following order (EV_MSC events): A, B, X, Y, LB, RB, Back, Menu, LS, RS, Xbox logo.

All good, except for the Xbox logo. Here's RS:
```
Event: time 1593865287.865560, type 4 (EV_MSC), code 4 (MSC_SCAN), value 9000a
Event: time 1593865287.865560, type 1 (EV_KEY), code 318 (BTN_THUMBR), value 0
```
and the xbox logo:
```
Event: time 1593865289.545394, type 4 (EV_MSC), code 4 (MSC_SCAN), value c0223
Event: time 1593865289.545394, type 1 (EV_KEY), code 172 (KEY_HOMEPAGE), value 0
```

but this is normal, isn't it?

> evtest should have correctly assigned the BTN_* symbols to each of these events.

Looks good to me! (obviously the paddles are still messy, but I don't think it's a goal for the first version)

> jstest orders the buttons in the same order but excludes the Xbox logo button.

Uhm, I do have an 11th button, appearing as `KEY_SHUFFLE`, but I cannot trigger it in any way. Otherwise good.

> jstest reports no blind buttons.

See above.

> HTML5 gamepad tester

Wow, you weren't kidding about browsers being messyl This is a screenshot of https://gamepadviewer.com/# (wanted something a bit more visual) while I'm holding the start button (left - Firefox, right - Brave)

image

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -1,4 +1,4 @@
-#define DRV_VER "@DO_NOT_CHANGE@"
+#define DRV_VER "v0.7-41-g8704a01-dirty"
```
ehats, line 1: huh, sneaky! I'll get rid of this once we're getting ready to merge

kakra: Yeah, I'm going to fix this in the next release cycle... The build processes modifies files in-place. It should not do this at all.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -726,6 +757,11 @@ static int xpadneo_input_mapping(struct hid_device *hdev, struct hid_input *hi,
 {
 	int i = 0;
 
+	/* XBE2 reports a full keyboard, which we don't need */
+	if ((usage->hid & HID_USAGE_PAGE) == HID_UP_KEYBOARD) {
+		return MAP_IGNORE;
```
kakra, line 762: According to Linux kernel code style guide, a single-instruction if body should not have curly braces around it.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -217,6 +217,37 @@ static const struct usage_map xpadneo_usage_maps[] = {
...
+	USAGE_IGN(0x90013), /* copy of X */
+	USAGE_IGN(0x90014), /* copy of Y */
+	USAGE_IGN(0x90016), /* copy of LB */
+	USAGE_IGN(0x90017), /* copy of RB */
+	USAGE_IGN(0x9001B), /* copy of Start */
+	USAGE_IGN(0x9001D), /* copy of LS */
+	USAGE_IGN(0x9001E), /* copy of RS */
+	USAGE_IGN(0xC0082), /* copy of Select button */
+
+	/* XBE2: Disable unused "TRIGGER_HAPPY" buttons */
```
kakra, line 244: You may need to map some of these to other symbols than `BTN_TRIGGER_HAPPY_*` when you're going to deal with the paddles, ideally these should get new `0x9....` codes through bit-shifting, directly following the `0x9000b` for the logo button. On a side note: I wonder how much beer went into the kernel development when naming the buttons. :-)

Due to ordering constraints, you can only use button symbols below `#define BTN_DIGI 0x140` (kernel source `input-event-codes.h`) otherwise the input layer will order them in front of `BTN_A` which we don't want (messes up `jstest`).

ehats:
Oops, I just tried this out with Terraria (had to find something that needs a lot of buttons and I have it installed) - good results overall, with the exception of LS and RS.

LS got detected as `BigButton` in-game, and RS was detected as LS.

`evtest` looks good to me overall, so I wanted to ask if you could double-check the jstest mappings:

```
Joystick (Xbox Elite Wireless Controller) has 8 axes (X, Y, Z, Rx, Ry, Rz, Hat0X, Hat0Y)
and 11 buttons (BtnA, BtnB, BtnX, BtnY, BtnTL, BtnTR, BtnSelect, BtnStart, BtnThumbL, BtnThumbR, (null)).
```

hid-xpadneo/src/hid-xpadneo.c
kakra: OTOH, there don't seem to be many other options than `BTN_TRIGGER_HAPPY*`.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -850,7 +888,7 @@ static int xpadneo_raw_event(struct hid_device *hdev, struct hid_report *report,
 	}
 
 	/* correct button mapping of Xbox controllers in Linux mode */
-	if (report->id == 1 && (reportsize == 17 || reportsize == 39)) {
+	if (report->id == 1 && (reportsize == 17 || reportsize == 39 || reportsize == 55)) {
```
kakra, line 891: You probably need to make a variant of this in a second if block for the paddle buttons. Ordering may be important if you check the same bytes, as we replace bytes in `data[]` in-place.

kakra:
> Oops, I just tried this out with Terraria (had to find something that needs a lot of buttons and I have it installed) - good results overall, with the exception of LS and RS.
> 
> LS got detected as `BigButton` in-game, and RS was detected as LS.
> 
> `evtest` looks good to me overall, so I wanted to ask if you could double-check the jstest mappings:
> 
> ```
> Joystick (Xbox Elite Wireless Controller) has 8 axes (X, Y, Z, Rx, Ry, Rz, Hat0X, Hat0Y)
> and 11 buttons (BtnA, BtnB, BtnX, BtnY, BtnTL, BtnTR, BtnSelect, BtnStart, BtnThumbL, BtnThumbR, (null)).
> ```
> 
> this looks good to my layman eyes

Yep, ignore this for now. It's probably the `(null)` button that orders before the thumb buttons in the game, probably through SDL. It may be gone when you get rid of this unnamed button. But this issue can also be introduced by SDL itself when it has a mapping entry for your controller signature (`BUS:VID:PID:VERSION`). There's a read-up for it here (slightly out of date, some of this doesn't apply to the current driver but as you're adding a new device, you may find yourself in a similar situation): https://github.com/atar-axis/xpadneo/blob/master/docs/SDL.md

This is why we may need to change the `PID` or `VERSION` field after you got the untranslated mappings into a sane state (untranslated aka "no SDL or browser API involved").

ehats:
Alright, so I'll try to get rid of that last unused button as it doesn't really make sense either way, and then I think we'll at least have every axis and button represented with no duplicates or anything weird (except for paddles, still)

kakra:
> Alright, so I'll try to get rid of that last unused button as it doesn't really make sense either way, and then I think we'll at least have every axis and button represented with no duplicates or anything weird (except for paddles, still)

I think you may need to just patch the PID from `0xB05` to `0x0B00` to pretend USB mode. The code for this is already prepared but disabled with `#if 0`. I'm not sure if SDL has a built-in mapping for `0x0B00`, it may be worth looking for that in the SDL source code. I don't really like the idea to prentend the wireless controller as wired but I don't see another option currently as we don't know the Bluetooth PID for the controller in Windows mode.

kakra:
The SDL mapping DB is here: http://hg.libsdl.org/SDL/file/default/src/joystick/SDL_gamecontrollerdb.h

kakra:
I just noticed that SDL now supports an environment variable `SDL_GAMECONTROLLER_USE_BUTTON_LABELS` which, when set to 1, prefers the input symbols instead of position when mapping controllers. So SDL games just work regardless of the ordering, it just needs correct symbol names (as shown in `jstest` or `evtest`). It may even default to 1, I didn't test this yet. But it may explain why some Proton games went crazy lately.

kakra:
> I think you may need to just patch the PID from `0x0B05` to `0x0B00` to pretend USB mode.

The SDL mapping DB doesn't seem to have an explicit mapping of `0x0B00` - so this may already fix things. It only has two `0x0B05` mappings, one for firmware version `0x0902` und one for `0x0903` (I think `0x09` is for the Linux version of the firmware only, Windows uses `0x0403` as far as I remember). So patching just the version number could fix SDL but the browsers do not look at the version, they only compare the PID.

ehats:
I tried changed the PID but didn't notice any changes anywhere - I tried both the 02E0 that was already there, as well as the 0B05. I didn't notice any changes in jstest, evtest or terraria whatsoever

kakra:
Changing the announced PID won't affect `jstest` or `evtest` at all, only layers that look at this, like SDL. Also programs using evdev or joydev directly don't care for this value.

ehats:
I managed to get rid of the last blind button, and I also disabled the paddles and the profile switcher since they're kinda just noise for now.

Where do we go from here?

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -739,10 +779,13 @@ static int xpadneo_input_mapping(struct hid_device *hdev, struct hid_input *hi,
 				hid_map_usage_clear(hi, usage, bit, max,
 						    entry->ev.event_type, entry->ev.input_code);
 			}
+			hid_info(hdev, "MAP: %lx -> %lx\n", usage->hid, entry->ev.input_code);
```
ehats, line 782: will be removed

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -1,4 +1,4 @@
-#define DRV_VER "@DO_NOT_CHANGE@"
```
kakra, line 1: Just a comment as a reminder to remove this...

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -758,9 +763,8 @@ static int xpadneo_input_mapping(struct hid_device *hdev, struct hid_input *hi,
 	int i = 0;
 
 	/* XBE2 reports a full keyboard, which we don't need */
-	if ((usage->hid & HID_USAGE_PAGE) == HID_UP_KEYBOARD) {
+	if ((usage->hid & HID_USAGE_PAGE) == HID_UP_KEYBOARD)
```
kakra, line 766: This may actually be part of chatpad support but I'm not sure if that is even supposed to work via Bluetooth. Unless someone has this hardware, we can't test it, and it should stay disabled as it is now.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -241,13 +241,18 @@ static const struct usage_map xpadneo_usage_maps[] = {
 	USAGE_IGN(0x9001E), /* copy of RS */
 	USAGE_IGN(0xC0082), /* copy of Select button */
 
-	/* XBE2: Disable unused "TRIGGER_HAPPY" buttons */
-	USAGE_IGN(0x90012),
+	/* XBE2: Disable extra features until proper support is implemented */
+	USAGE_IGN(0xC0081), /* Four paddles */
+	USAGE_IGN(0xC0085), /* Profile switcher */
```
kakra, line 246: The profile switcher exposes itself as a HID event? Than implementation may become more tricky as we may need to hook-up some state change to this. Nevertheless, it looks kinda redundant to me, as we also get the profile ID in the raw report data.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -241,13 +241,18 @@ static const struct usage_map xpadneo_usage_maps[] = {
 	USAGE_IGN(0x9001E), /* copy of RS */
 	USAGE_IGN(0xC0082), /* copy of Select button */
 
-	/* XBE2: Disable unused "TRIGGER_HAPPY" buttons */
-	USAGE_IGN(0x90012),
+	/* XBE2: Disable extra features until proper support is implemented */
```
kakra, line 244: If you want to get this merged in this state, you should add this to the docs, too: Put some information that the extra buttons and functions are not supported yet, and that only profile 0 has been tested.

ehats:
Thanks for the feedback! I'll clean up the indentation, remove the debug leftovers, update the docs with the caveats, update the PID and squash the commits, hopefully tomorrow!

> you could pair the controller with Windows 

I don't have access to a Windows device - I could probably download an evaluation image and spin up a VM, but I'm not sure if that'd work at all

kakra:
> I don't have access to a Windows device - I could probably download an evaluation image and spin up a VM, but I'm not sure if that'd work at all

Maybe just ask a friend... Otherwise: Pass-through for USB Bluetooth adapters works without problems in VMs.

hid-xpadneo/src/hid-xpadneo.c
ehats: I tried to find the user-facing controller-specific sections, but I'm not sure there is one? I can add a new section to the readme, or put these notes next to the descriptor. Which one would you prefer?

ehats:
> I'll clean up the indentation, remove the debug leftovers, update the docs with the caveats, update the PID and squash the commits, hopefully tomorrow!

All done, except documenting the caveats for now (see my question above)

hid-xpadneo/src/hid-xpadneo.c
kakra: I think it should be in the `README.md`, somewhere near the feature description.

docs/descriptors/xbe2_unknown.md
```diff
@@ -0,0 +1,690 @@
...
+0x09, 0x32,        //     Usage (Z)  // Fixed up to RX
+0x09, 0x35,        //     Usage (Rz) // Fixed up to Ry
+0x15, 0x00,        //     Logical Minimum (0)
+0x27, 0xFF, 0xFF, 0x00, 0x00,  //     Logical Maximum (65534)
+0x95, 0x02,        //     Report Count (2)
+0x75, 0x10,        //     Report Size (16)
+0x81, 0x02,        //     Input (Data,Var,Abs,No Wrap,Linear,Preferred State,No Null Position)
+0xC0,              //   End Collection
+0x05, 0x02,        //   Usage Page (Sim Ctrls)
+0x09, 0xC5,        //   Usage (Brake) // Fixed up to Z (LT)
```
kakra, line 140: Isn't this a comment you wanted to remove? Tho, it may make sense to have it here.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -1070,14 +1106,14 @@ static void xpadneo_remove(struct hid_device *hdev)
 
 static const struct hid_device_id xpadneo_devices[] = {
 	/* XBOX ONE S / X */
-	{HID_BLUETOOTH_DEVICE(USB_VENDOR_ID_MICROSOFT, 0x02FD)},
-	{HID_BLUETOOTH_DEVICE(USB_VENDOR_ID_MICROSOFT, 0x02E0)},
+	{ HID_BLUETOOTH_DEVICE(USB_VENDOR_ID_MICROSOFT, 0x02FD) },
```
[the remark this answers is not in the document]
ehats: yup, these got here after running `indent`. will undo

docs/descriptors/xbe2_unknown.md
ehats: It used to just say `FIXED` and I was going to remove it, but I found myself looking at these a lot, so I figured I'd expand it a bit, but keep it around.

kakra: Fine for me.

ehats:
Got rid of the unrelated changes and rebased

hid-xpadneo/src/hid-xpadneo.c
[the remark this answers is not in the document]
kakra: Did you run `indent` from the project root?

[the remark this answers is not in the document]
ehats: I ran `indent hid-xpadneo/src/hid-xpadneo.c` from the project root

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -892,16 +932,12 @@ static int xpadneo_input_configured(struct hid_device *hdev, struct hid_input *h
 	 * 0xB05 wireless Linux mode (Android mode)
 	 */
 	switch (xdata->idev->id.product) {
-#if 0
 	case 0x0B05:
-		/* FIXME What's the Windows ID of this controller? */
 		hid_info(hdev,
-			 "pretending XBE2 Windows wireless mode (changed PID from 0x%04X to 0x???)\n",
+			 "pretending XBE2 wireless Linux mode (changed PID from 0x%04X to 0x0B00)\n",
```
kakra, line 937: This should read "pretending XBE2 wired mode" as this is what it does: `0x0B00` is the wired USB ID of the controller.

ehats:
Fixed the comment, but just realised something - does this PR really solve the original issue? This is a good start, but it's not quite solved, is it?

If you agree, I'll remove that bit from the commit message and keep the issue open

ehats:
Okay, played around with a couple more games, but none of them would work perfectly. Even though evtest and jstest do everything as I'd expect, native games are having issues with LS and RS. LS isn't really being recognised, and RS is being recognised as LS.

Wine/proton games were even messier, with a bunch of buttons not being where I'd expect (e.g. X doesn't work, Y appear as X, LB appears as Y)

Any idea where things are going wrong if jstest and evtest are seemingly normal?

kakra:
Okay, then it looks like SDL messes with our mappings. For a proper fixup we need to know how the HTML5 gamepad API behaves.

kakra:
It's strange because `045e:0b00` is not in the SDL database, so it should work just fine.

kakra:
> Okay, played around with a couple more games, but none of them would work perfectly. Even though evtest and jstest do everything as I'd expect, native games are having issues with LS and RS. LS isn't really being recognised, and RS is being recognised as LS.

This is usually the result when the Xbox logo button still exists in the `jstest` output... Do you have a custom button mapping in `/var/lib/joystick`?
 
> Wine/proton games were even messier, with a bunch of buttons not being where I'd expect (e.g. X doesn't work, Y appear as X, LB appears as Y)

This looks like SDL tried to correct the mappings that our driver already corrected. But it's not in the SDL as `0x0B00` so it should not do that. My current best guess is that Wine itself tries to remap this controller. I'll look at the source code, it has it's own device database.

> Any idea where things are going wrong if jstest and evtest are seemingly normal?

Please show the output of both.

ehats:
Alright, the wine issue may have been due to some long-running wine processes. I killed everything, tried again, and now wine games are also in the same state: almost working, except for the LS-RS issue. 

> Do you have a custom button mapping in /var/lib/joystick?
There's no such file or folder on my machine, so I guess no? I did check what `jscal` has to offer and found the "current mapping":

```
~ ❯❯❯ jscal -q /dev/input/js0                                                                      
jscal -u 8,0,1,2,3,4,5,16,17,10,304,305,307,308,310,311,314,315,317,318 /dev/input/js0
```

jstest output:

```
jstest /dev/input/js0
Driver version is 2.1.0.
Joystick (Xbox Elite Wireless Controller) has 8 axes (X, Y, Z, Rx, Ry, Rz, Hat0X, Hat0Y)
and 10 buttons (BtnA, BtnB, BtnX, BtnY, BtnTL, BtnTR, BtnSelect, BtnStart, BtnThumbL, BtnThumbR).
```

pressing the Xbox button doesn't cause anything to be printed in jstest.

evtest output:

[evtest.txt](https://github.com/atar-axis/xpadneo/files/4875529/evtest.txt)

Pressing and releasing the xbox button causes the following events:
```
Event: time 1593980704.890120, type 4 (EV_MSC), code 4 (MSC_SCAN), value c0223
Event: time 1593980704.890120, type 1 (EV_KEY), code 172 (KEY_HOMEPAGE), value 1
Event: time 1593980704.890120, -------------- SYN_REPORT ------------
Event: time 1593980705.009132, type 4 (EV_MSC), code 4 (MSC_SCAN), value c0223
Event: time 1593980705.009132, type 1 (EV_KEY), code 172 (KEY_HOMEPAGE), value 0
Event: time 1593980705.009132, -------------- SYN_REPORT ------------
```

kakra:
`jscal -q` reports `8` axis positions and all look fine, then `10` buttons with the Linux input keycodes, starting with 304 for `BTN_A`, then 310 for `BTN_TL` (which is the button above the trigger), 314/315 is `BTN_SELECT`/`BTN_START`, followed by 317/318 for the thumb stick buttons. This looks all good.

I also looked at the wine code. There's a bus driver which exposes SDL devices on the bus as HID devices. This works by crafting a custom HID report descriptor: https://github.com/ValveSoftware/wine/blob/proton_5.0/dlls/winebus.sys/bus_sdl.c

This SDL bus is still fairly new but there seems to be something even newer: Instead of setting the button count to 15 and adjusting it later, there's some `CONTROLLER_BUTTONS_XINPUT_HACK`, and it now has two byte arrays, one for 10 buttons and one for the hacked version. I'll need to dig into that to see what's going on but I suppose that is what's messing our mapping up.

Actually, I think what you're seeing now is what I saw previously: I exposed 11 buttons in xpadneo and all of a sudden, Proton games were messed up. So I changed that to 10 buttons and now it works. I think this is what happens: My controller wasn't picked up by the hack but yours is. So we just need to find out when Proton does apply the hack and work around that.

kakra:
Seems to be somewhere around here: https://github.com/ValveSoftware/wine/blob/2409bd1f74be116172688a25df725290637c255a/dlls/winebus.sys/bus_sdl.c#L920

ehats:
Just to clarify: we're talking about proton, but SDL is the key here, right? Since the LS/RS problem affects both wine games, any native ones alike :thinking:

kakra:
From looking at the Proton code, it should map your controller just right as long as SDL does the correct thing. So I think we have to look at SDL. Although I'm not sure what the `xinput_hack` this is supposed to do. It's somehow attached to the existence of a guide button but it looks like this is a hack to actually remove the guide button from the HID report as the Windows native reports don't have this button. So basically the same thing we do in xpadneo.

Now let's find out why your SDL is confused about the buttons. Do you have any environment variable set beginning with `SDL_*`?

ehats:
`env | grep -i SDL` gets absolutely nothing, so it doesn't seem so. I also don't remember ever fiddling with it

kakra:
You could try https://www.generalarcade.com/gamepadtool/linux/

I just downloaded the zip, unpacked it to some directory and ran the binary. It should be quite helpful as it also seems to show if a mapping is used.

kakra:
Okay, I've found my controller to have a mapping entry:

https://hg.libsdl.org/SDL/file/default/src/joystick/SDL_gamecontrollerdb.h line 373 (reordered for better readability):
```
"050000005e040000e002000003090000,Xbox One Wireless Controller,
a:b0,b:b1,
x:b2,y:b3,
leftshoulder:b4,
rightshoulder:b5,
back:b6,
start:b7,
leftstick:b8,
rightstick:b9,
guide:b10,
dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,
leftx:a0,lefty:a1,
lefttrigger:a2,
rightx:a3,righty:a4,
righttrigger:a5,
",
```

So this is actually a 1:1 mapping: Each b-position corresponds to xpadneo mapping 1:1.

But your xpadneo is pretending to be `050000005e040000000b000003090000` which does not exist in the database, so it should map 1:1 also.

Just to cross-check: `jstest` and `evtest` both report all the buttons correct for you, do they?

ehats:
Yes, both of them look correct to me

kakra:
Could you try to run Steam or SDL games with the following variable set:
```bash
export SDL_CONTROLLERCONFIG='050000005e040000000b000003090000,Xbox One Wireless Controller,\
a:b0,b:b1,back:b6,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b10,leftshoulder:b4,leftstick:b8,\
lefttrigger:a2,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b9,righttrigger:a5,rightx:a3,righty:a4,start:b7,\
x:b2,y:b3,'
```

ehats:
I got that tool, this is what it says:
```
Searching gamepads...
Found 1 gamepad(s):
"XInput Controller", 050000005e040000000b000003090000 (mapping available)
Environment variable "SDL_GAMECONTROLLERCONFIG" is not defined
Checking if new mappings available from github: https://github.com/gabomdq/SDL_GameControllerDB
Mapping string for "XInput Controller" is "(null)"
```

kakra:
> I got that tool, this is what it says:

So it synthesizes a mapping because there is no such mapping in the SDL database. This may be the new feature introduced in SDL to build a mapping from the HID description in Linux. Then something is going wrong in SDL with this feature.

Maybe just patch the PID to `0x02E0` for now by just dropping the `case` and adjusting the one below:
```c
	case 0x02FD:
	case 0x0B05:
```

Does it work then? Otherwise we have to look at the SDL source and see what its magic is doing.

ehats:
Made the changes and re-tested with that tool:
```
Searching gamepads...
Found 1 gamepad(s):
"Xbox One Wireless Controller", 050000005e040000e002000003090000 (mapping available)
Environment variable "SDL_GAMECONTROLLERCONFIG" is not defined
Checking if new mappings available from github: https://github.com/gabomdq/SDL_GameControllerDB
Mapping string for "Xbox One Wireless Controller" is "050000005e040000e002000003090000,Xbox One Wireless Controller,a:b0,b:b1,back:b6,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b10,leftshoulder:b4,leftstick:b8,lefttrigger:a2,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b9,righttrigger:a5,rightx:a3,righty:a4,start:b7,x:b2,y:b3,"
```

kakra:
> Made the changes and re-tested with that tool:

Do games now work correctly?

kakra:
I wonder if the bit-shifting in `xpadneo_raw_report` works correctly for you, as you wrote:

> Pressing and releasing the xbox button causes the following events:
> 
> ```
> Event: time 1593980704.890120, type 4 (EV_MSC), code 4 (MSC_SCAN), value c0223
> Event: time 1593980704.890120, type 1 (EV_KEY), code 172 (KEY_HOMEPAGE), value 1
> Event: time 1593980704.890120, -------------- SYN_REPORT ------------
> Event: time 1593980705.009132, type 4 (EV_MSC), code 4 (MSC_SCAN), value c0223
> Event: time 1593980705.009132, type 1 (EV_KEY), code 172 (KEY_HOMEPAGE), value 0
> Event: time 1593980705.009132, -------------- SYN_REPORT ------------
> ```

If the bit is set for the Xbox logo button, it should show `MSC_SCAN`  as `9000b`.

OTOH, it may just send report ID 2. My controller doesn't although it is defined in the descriptor. So my event comes from report ID 1 (which has the button bits) and yours comes from report ID 2 (which has a single bit for "AC Home").

kakra:
Quick note: Profile 0 doesn't enable the paddles anyways: https://www.youtube.com/watch?v=rGxLMjErCJU

kakra:
I prepared profile switching in my patch queue which you may want to make use of in the next release cycle. I'm going to add a sysfs interface which exposes all four profiles to adjust parameters like profile switching. We could even create a GUI app for that then.

ehats:
I downloaded the windows version of the gamepad tool as well, ran it through wine and all the buttons were recognised correctly (except for the Xbox logo, but I believe that's normal?). Same for the Linux version, and I didn't notice any issues with the few games I tried either.

ehats:
This is what `btmon` reports for the xbox logo press-release cycle:

```
> ACL Data RX: Handle 256 flags 0x02 dlen 7                                                                                                                                                   #24 [hci0] 7.145637
      Channel: 65 len 3 [PSM 0 mode Basic (0x00)] {chan 65535}
        a1 02 01                                         ...
> ACL Data RX: Handle 256 flags 0x02 dlen 7                                                                                                                                                   #25 [hci0] 7.335255
      Channel: 65 len 3 [PSM 0 mode Basic (0x00)] {chan 65535}
        a1 02 00 
```

in case that helps with anything

kakra:
So I believe patching the PID is the way to go for v0.8 then, and we have to revisit this later when you add the paddles.

> This is what `btmon` reports for the xbox logo press-release cycle:

Yeah, this is what I suspected. Thanks for confirming.

So I'll wait for the last force-push but will not merge before this evening (busy work day today).

If you want to start working on the profile stuff using my patch from my patch queue, you should open a new PR for it.

ehats:
Aaand there's the force push :) I'll take a look at the profile stuff a bit later, I'm guessing I'll probably be listening for a different event. Will play around and open a PR.

Thanks for all the help on this one! I knew literally nothing about HID a week ago ;)

kakra:
> I downloaded the windows version of the gamepad tool as well, ran it through wine and all the buttons were recognised correctly (except for the Xbox logo, but I believe that's normal?). Same for the Linux version, and I didn't notice any issues with the few games I tried either.

Yes that's normal. But didn't we need to patch the PID to `0x02E0`? It seems you left it at `0x0B00` and everything works now?

ehats:
Oh wait, I always tested it with `0x02E0`, but apparently never pushed it to github :facepalm:

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -1,4 +1,4 @@
-#define DRV_VER "@DO_NOT_CHANGE@"
+#define DRV_VER "v0.7-42-g22b75ba-dirty"
```
kakra, line 1: Whoops...

kakra: BTW: I'm going to fix this dumb behavior today as it causes more problems: #217

kakra:
Final. Thanks a lot for your efforts and endurance, also for taking the opportunity of learning something new.

kakra:
@ehats A user donated an XBE2 controller to me, so I can build upon your work now. I think we'll need to revert some filtering, the controller should be exposed as three different sub devices: game input, keyboard, and consumer control. I was having that idea before, so I'll queue that for 0.10. But I'll add support for the paddle buttons in 0.9 at least. Also, some HID event filtering is missing, I'll also add that. And we can detect switching the trigger limiters, so I'll also add that to the driver (without making any use of it, just preparing stuff).

Issue #228, opened 2020-07-12
'Y' and 'R' button not detected by SDL applications


terencode:
## Version of xpadneo
v0.8.r515.g715c2e9

## Describe the bug
When connecting an 8bitdo controller (tested with both the zero 2 and SN30 pro+) in X Input Mode over bluetooth,
the keys labeled 'Y' and 'R' are not registered in sdl controllermap and antimicrox.

## Steps to Reproduce

1. Connect an 8bitdo zero 2 controller over bluetooth in X Input Mode (Start + X)
2. Run jstest/jstest-gtk and notice all buttons are correctly registered
3. Run antimicrox or sdl controllermap and notice the 'Y' and 'R' buttons do not do anything

## Expected behavior

All the buttons should be detected.

## System information

```console
# uname -a
Linux terence-desktop 5.7.8-18-tkg-bmq #1 TKG SMP PREEMPT Fri, 10 Jul 2020 18:00:53 +0000 x86_64 GNU/Linux
```

```console
# xxd -c20 -g1 /sys/module/hid_xpadneo/drivers/hid:xpadneo/0005:045E:*/report_descriptor | tee >(cksum)
00000000: 05 01 09 05 a1 01 85 01 09 01 a1 00 09 30 09 31 15 00 27 ff  .............0.1..'.
00000014: ff 00 00 95 02 75 10 81 02 c0 09 01 a1 00 09 33 09 34 15 00  .....u.........3.4..
00000028: 27 ff ff 00 00 95 02 75 10 81 02 c0 05 01 09 32 15 00 26 ff  '......u.......2..&.
0000003c: 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01 81 03 05 01 09  ...u.....%.u........
00000050: 35 15 00 26 ff 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01  5..&....u.....%.u...
00000064: 81 03 05 01 09 39 15 01 25 08 35 00 46 3b 01 66 14 00 75 04  .....9..%.5.F;.f..u.
00000078: 95 01 81 42 75 04 95 01 15 00 25 00 35 00 45 00 65 00 81 03  ...Bu.....%.5.E.e...
0000008c: 05 09 19 01 29 0a 15 00 25 01 75 01 95 0a 81 02 15 00 25 00  ....)...%.u.......%.
[...]

PR #229, opened 2020-07-12
Fix 8BitDo controllers layout and button mapping

kakra:
8BitDo controllers work in native Windows reporting mode, so we should not try to fixup the button bits. Also, these controllers have a Nintendo button layout which swaps A with B and X with Y compared to Xbox controllers.


kakra:
Affects #228 (button order is correct now in `jstest` and `evtest`) but does not ultimately fix it. Merged without closing the issue.

PR #232, opened 2020-07-17
Rework dead zone handling

kakra:
Evidence shows that there is a problematic code path between the Linux kernel and games when using Wine/Proton. Windows games do not expect the input drivers to have dead zones applied but Linux drivers imply that at every level. Since Wine/Proton uses SDL for creating a bus device, it will see a game controller with dead zones already handled - which games do not expect.

This series of commits shrinks the default dead zones and fuzz filter for more precision and adds a kill switch for disabling handling of dead zones completely so Wine/Proton games will have their expectations fulfilled.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -110,7 +110,7 @@ function parse_args {
     # If line doesn't exist echo all of the defaults.
     mkdir -p "$(dirname "${CONF_FILE}")"
     touch "${CONF_FILE}"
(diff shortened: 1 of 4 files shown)


kakra:
@a-kring Please test.

kakra:
Next SDL2 version is going to fix this, Proton games will properly pick this up then as soon as Proton ships the new version: https://bugzilla.libsdl.org/show_bug.cgi?id=5241#c17

Issue #247, opened 2020-10-27
Xbox Elite Series 2 intermittently but consistently locks up


solystm:
## Version of xpadneo
v0.8
(I also have xow 0.5-17-gaf5b9d4 installed, but the behavior is the same with xpadneo alone, xow alone, and xpadneo with xow)

## Severity / Impact
It completely prevents the controller from being used through Bluetooth for more than between 5 minutes and an hour, depending on some factors I'm not sure on. After that period it locks up and needs to be powered off and then reconnected again. Games generally need to be restarted in order to re-detect it.

* [x] I've read the docs and the bug reporting instructions
* [ ] It does not work at all
* [ ] It used to work in a previous version
* [x] It mostly works but sometimes it doesn't
* [x] I found a work-around
* [x] I probably didn't figure it all out but it's too early to give up
* [ ] I don't know how to ... 
* [ ] It's too complicated
* [ ] Fantastic work but ... 
* [ ] I can code and I want to help

## Describe the bug
The controller connects fine and works well, reporting battery, all buttons working, even the underpad triggers (though that may be xow's doing?) I start playing a game (tested with Trials of Mana and Streets of Rage 4 primarily) and it's detected and works as expected in the game. However, after a period of time somewhere between 5 minutes (at the earliest) and about an hour (at the latest) the controller will lock up and stop working until it's powered off (hold Xbox button) and then powered back on again.

By "lock up" I mean the controller's input 'sticks' on whatever it was doing when the bug occurred. So, for example, if i was holding right on the left analog, it would continue holding right even if i let go entirely. Other buttons cease to register at this time. For example, pressing ABXY gets no input in game or in a controller calibration utility, when both work fine normally. The controller calibration utility will report the last state (for example, the left analog being held to the right) and no longer update. The controller stays connected to bluetooth from what i can see.

## Steps to Reproduce
Connect up, start a game, and play. Eventually, it will drop the connection. With that said, it seems to be associated more with left analog stick inputs. It's never actually locked up while the left analog was neutral. That includes an extended session of Streets of Rage 4 only using the dpad, but when I switched to the left analog it locked up pretty much immediately. Given the wide range in the amount of time that passes it's difficult to say if it's related but it sure seems that way.

## Expected behavior
It should continue to work normally without doing that :)

## Screenshots/Gifs
I could take a video of me giving the pad inputs while nothing is registered by a game/controller calibration tool, but I'm not sure that would be terribly useful.

## System information
```console
# uname -a
[...]

Issue #243, opened 2020-09-08
Xbox Series 2 Controller input stops during rumble if attenuation isn't disabled


schmitmd:
## Version of xpadneo

Latest master branch, commit 041e8d8cc9a1ec3184a12fe057b8d7d568e72f23

## Severity / Impact

The controller is essentially unusable if the workaround is not employed.
* [x] I've read the docs and the bug reporting instructions
* [ ] It does not work at all
* [ ] It used to work in a previous version
* [ ] It mostly works but sometimes it doesn't
* [x] I found a work-around
* [x] I probably didn't figure it all out but it's too early to give up
* [ ] I don't know how to ... 
* [ ] It's too complicated
* [ ] Fantastic work but ... 
* [ ] I can code and I want to help

## Describe the bug

After game startup, if rumble is encountered, the controller stops responding and often gets "stuck" sending the same input as whatever was last pressed.
## Steps to Reproduce

With an Xbox Elite Series 2 controller bound via Bluetooth (no USB dongle in my case), begin any game and play for a minute or two until the controller rumbles, most noticeably while holding the left trigger.  New input will cease to work and there will often be a continuation of whatever the last controller state was (i.e. if you were holding the left trigger, the input will seem as though you never let go).
## Expected behavior

The controller rumbles and continues with normal game input.
## System information

```console
Linux myhost 5.8.7-arch1-1 #1 SMP PREEMPT Sat, 05 Sep 2020 12:31:32 +0000 x86_64 GNU/Linux
```

[Link to output](https://gist.github.com/schmitmd/ff2e1f2050c07698deea5fcc347488cb) of `xxd -c20 -g1 /sys/module/hid_xpadneo/drivers/hid:xpadneo/0005:045E:*/report_descriptor | tee >(cksum)`
## Controller and Bluetooth information

[Link to output](https://gist.github.com/schmitmd/0fb6cc4d983807363fd1c2a6d520472f) of `sudo btmon | tee xpadneo-btmon.txt`

[Link to output](https://gist.github.com/schmitmd/2374f88e5302b54d8a78dd8f79b945d9) of `dmesg | egrep -i 'hid|input|xpadneo' | tee xpadneo-dmesg.txt`

There is no USB dongle involved, the pairing is via Bluetooth connection to the laptop's `8087:0aaa Intel Corp. Bluetooth 9460/9560 Jefferson Peak (JfP)` Bluetooth chip.  

## Additional context

[...]

Issue #198, opened 2020-06-02
Bluetooth connection stability broken after kernel update


kakra:
This is a information collection issue to pinpoint the kernel change that broke the Bluetooth stability with the controller. I'll ask anyone who wants to help to provide the following information:

**Please provide kernel versions as shown by `uname -a` if possible because distributions often implement their own patchlevel versioning.**

1. Which distribution are you running?
2. Last known working kernel version
3. First known broken kernel version
4. Output of `lsusb` and take note of the device number of your USB dongle
5. File generated with `lsusb -v -s## | tee xpadneo-lsusb.txt` where `##` is the device number picked in the previous step
6. Disconnect your controller, run `sudo btmon | tee xpadneo-btmon.txt` and connect the controller, post the resulting file.

Thanks!

envyniv:
1. Ubuntu
2. I'm not sure, but i think any 5.3?
3. current; 5.4.0-33-generic
4/5. [xpadneo-lsusb.txt](https://github.com/atar-axis/xpadneo/files/4722057/xpadneo-lsusb.txt)
6. [xpadneo-btmon.txt](https://github.com/atar-axis/xpadneo/files/4722058/xpadneo-btmon.txt)
hope i managed to help in some way. :)

envyniv:
i know i shouldn't chat here, but am i the only one that seeme to corelate this issue with the kernel update?

kakra:
I think this will get some attention with the next version bump.

kakra:
> So ,will next version support xbox-one-s? ( it just couldn't connected pc(ubuntu-1804) through bluetooth)

I don't want to be rude but this issue is solely about collecting information with regressions. Please look at the other issues, the one's linked above are a good starting point. Otherwise feel free to open a new issue report. I'll mark off-topic comments as off-topic now. Thanks in advance for understanding and your cooperation.

kakra:
## HEADS UP

Anyone affected may tray #204 and report in #189 or #180.

envyniv:
> ## HEADS UP
> 
[...]

PR #253, opened 2020-11-08
Implement a rumble delay pool

kakra:
Research showed that there are users who only experience the controller crash with longer bursts of rumble. This PR tries to implement a delay pool to work around bursts.

My personal tests show no positive effect, tho.

Affects: #243 
Affects: #247 
Maybe-related: #198 
See-also: https://answers.microsoft.com/en-us/xbox/forum/all/xbox-elite-series-2-disconnecting-from-pc/7e5a964a-16cc-4e83-8dec-aa8c858b875c


kakra:
This doesn't help stability and only increases complexity and rumble lagging. Closing.

Issue #239, opened 2020-08-19
Pairing fails silently (?)


N00byKing:
## Version of xpadneo

Reported Version is v0.8-26-g041e8d8

## Severity / Impact

* [x] I've read the docs and the bug reporting instructions
* [x] It does not work at all
* [ ] It used to work in a previous version
* [ ] It mostly works but sometimes it doesn't
* [ ] I found a work-around
* [ ] I probably didn't figure it all out but it's too early to give up
* [ ] I don't know how to ... 
* [ ] It's too complicated
* [ ] Fantastic work but ... 
* [ ] I can code and I want to help

## Describe the bug

Following the given Instructions, when trying to pair it will report success, though the Controller never leaves pairing mode.
If I still try to connect afterwards, I am greeted with the following:
```
Attempting to connect to 98:7A:14:98:A8:AD
Failed to connect: org.bluez.Error.NotAvailable
```
Trying to let the Controller auto-connect fails as well.
An Interesting note is that the KDE Bluetooth Applet reports that the Controller is connected.

Full `bluetoothctl` Output:
```
[bluetooth]# scan on
Discovery started
[NEW] Device 98:7A:14:98:A8:AD Xbox Wireless Controller
[bluetooth]# pair 98:7A:14:98:A8:AD
Attempting to pair with 98:7A:14:98:A8:AD
[CHG] Device 98:7A:14:98:A8:AD Connected: yes
[Xbox Wireless Controller]# trust
[CHG] Device 98:7A:14:98:A8:AD Trusted: yes
Changing 98:7A:14:98:A8:AD trust succeeded
[CHG] Device 98:7A:14:98:A8:AD ServicesResolved: yes
[CHG] Device 98:7A:14:98:A8:AD Paired: yes
Pairing successful
[Xbox Wireless Controller]# connect
Missing dev argument
[Xbox Wireless Controller]# connect 98:7A:14:98:A8:AD
Attempting to connect to 98:7A:14:98:A8:AD
Failed to connect: org.bluez.Error.NotAvailable
```

## System information

```console
# uname -a
Linux doggo-pc 5.8.0-2-MANJARO #1 SMP PREEMPT Sat Aug 8 17:55:27 UTC 2020 x86_64 GNU/Linux
```

```console
# xxd -c20 -g1 /sys/module/hid_xpadneo/drivers/hid:xpadneo/0005:045E:*/report_descriptor | tee >(cksum)
[...]

Issue #166, opened 2020-03-05
Can't connect xbox controller, (keeps disconnecting and connecting)


felixsanz:
I have the *common* problem where the controller keeps doing this:

```
[CHG] Device 5C:BA:37:81:C3:DC Connected: yes
[CHG] Device 5C:BA:37:81:C3:DC Connected: no
[CHG] Device 5C:BA:37:81:C3:DC Connected: yes
[CHG] Device 5C:BA:37:81:C3:DC Connected: no
```

I have tried everything for many months, so now i'm finally asking for some help :(

`disable_ertm` is set to Y (I've tried N too, same result). I'm on 5.5.7 kernel but i've tried many kernels and some kernel modules for my rtl8822be chip.

One big problem is that I can't re-pair the device, whenever i type `remove`, the controller is automatically paired again!

```
[bluetooth]# remove 5C:BA:37:81:C3:DC
[DEL] Device 5C:BA:37:81:C3:DC Xbox Wireless Controller
Device has been removed
[NEW] Device 5C:BA:37:81:C3:DC Xbox Wireless Controller
[CHG] Device 5C:BA:37:81:C3:DC Modalias: usb:v045Ep02E0d0903
[CHG] Device 5C:BA:37:81:C3:DC UUIDs: 00001124-0000-1000-8000-00805f9b34fb
[CHG] Device 5C:BA:37:81:C3:DC UUIDs: 00001200-0000-1000-8000-00805f9b34fb
[CHG] Device 5C:BA:37:81:C3:DC ServicesResolved: yes
[CHG] Device 5C:BA:37:81:C3:DC Paired: yes
[CHG] Device 5C:BA:37:81:C3:DC ServicesResolved: no
[CHG] Device 5C:BA:37:81:C3:DC Connected: no
[CHG] Device 5C:BA:37:81:C3:DC Connected: yes
[CHG] Device 5C:BA:37:81:C3:DC Connected: no
[CHG] Device 5C:BA:37:81:C3:DC Connected: yes
[CHG] Device 5C:BA:37:81:C3:DC Connected: no
[CHG] Device 5C:BA:37:81:C3:DC Connected: yes
[CHG] Device 5C:BA:37:81:C3:DC Connected: no
[CHG] Device 5C:BA:37:81:C3:DC Connected: yes
```

If i type `pair`, i get this:

```
[bluetooth]# pair 5C:BA:37:81:C3:DC
Attempting to pair with 5C:BA:37:81:C3:DC
[...]

PR #255, opened 2020-11-10
hid-xpadneo: Series X / S controller

lights0123:
I'm unable to test this as I experience #166 and #239, but I do get
```
[ 1103.799796] xpadneo 0005:045E:0B13.000E: fixing up Rx axis
[ 1103.799800] xpadneo 0005:045E:0B13.000E: fixing up Ry axis
[ 1103.799801] xpadneo 0005:045E:0B13.000E: fixing up Z axis
[ 1103.799802] xpadneo 0005:045E:0B13.000E: fixing up Rz axis
[ 1103.799803] xpadneo 0005:045E:0B13.000E: fixing up button mapping
[ 1103.800051] xpadneo 0005:045E:0B13.000E: enabling compliance with Linux Gamepad Specification
[ 1103.800097] input: Xbox Wireless Controller as /devices/virtual/misc/uhid/0005:045E:0B13.000E/input/input39
[ 1103.800221] xpadneo 0005:045E:0B13.000E: input,hidraw0: BLUETOOTH HID v5.01 Gamepad [Xbox Wireless Controller] on 44:16:22:ad:dr:es
[ 1103.800224] xpadneo 0005:045E:0B13.000E: controller quirks: 0x00000010
[ 1104.790606] xpadneo 0005:045E:0B13.000E: Xbox Wireless Controller [44:16:22:ad:dr:es] connected
```
in my dmesg logs, and I feel the pairing vibration pattern, although it doesn't send any further data.


kakra:
Oh cool, is this the original Xbox Series X controller? I didn't expect it to have Bluetooth... Usually the original controllers bundled with the consoles do not support Bluetooth.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -1215,6 +1215,9 @@ static const struct hid_device_id xpadneo_devices[] = {
 	{HID_BLUETOOTH_DEVICE(USB_VENDOR_ID_MICROSOFT, 0x02FD)},
 	{HID_BLUETOOTH_DEVICE(USB_VENDOR_ID_MICROSOFT, 0x02E0)},
 
+	/* XBOX ONE Series X / S */
+	{HID_BLUETOOTH_DEVICE(USB_VENDOR_ID_MICROSOFT, 0x0B13)},
```
kakra, line 1219: Looks fine but I'd rather move it below `0x0B05`.

kakra (commented):
There's more to it than just adding the PID to the list of supported devices, tho that's the first and most important step. We'll probably also have to pretend a different device model in `xpadneo_input_configured()`.

Also, please dump the device descriptor and document it in `docs/descriptors/`. You can use any of the other documents there as a template to get started. Just remove the section which you're unable to fill, a good starter could be the XBE2 documents.

Also, `dmesg` shows that the device descriptor can be parsed, and that the rumble initialized. So we should be able to get the buttons working: Use `evtest` to test if each button maps correctly, please also include the share button so we know the code of it. It should be added as an ignored button for now to the structure at the top of the file.

lights0123:
> Oh cool, is this the original Xbox Series X controller?

Yep, although I just got the controller by itself because a) all the other controllers were out of stock and b) might as well get the latest one—I don't actually own an Xbox, or plan to.

> Also, please dump the device descriptor and document it in `docs/descriptors/`.

Done, although I don't know what I should call it.

> Use evtest to test if each button maps correctly

The "change view" button (left button on face, I think also called Back/Select) simply had no output in `evtest`. In `btmon`, however, I got:
```
> ACL Data RX: Handle 3585 flags 0x02 dlen 23
      ATT: Handle Value Notification (0x1b) len 18
        Handle: 0x001e
          Data: 857d887d887e06840000000000000000
> ACL Data RX: Handle 3585 flags 0x02 dlen 23
      ATT: Handle Value Notification (0x1b) len 18
        Handle: 0x001e
          Data: 857d887d887e06840000000000000400
```
The second one (with the `4`) had the button pressed.

Select works. Y appears as West. B appears as East. X appears as North. A appears as South. Assuming that these are supposed to match with a compass, Y and X are swapped.

The DPad gives a positive value for down or right, and negative for up or left. Up and down are on the Y axis, left and right are on the X axis. This may mean that Y needs to be inverted.

The left joystick (ABS_X/ABS_Y) gives a positive value for up or right, and negative for down or left. Up and down are on the X axis, left and right are on the Y axis. This may mean that X and Y need to be swapped.
The right joystick (ABS_RX/ABS_RY) has the same behavior.

Left shoulder button sends BTN_TL, right shoulder button sends BTN_TR.

Left trigger (the variable one) sends:
```
Event: time 1605054206.497768, type 3 (EV_ABS), code 2 (ABS_Z), value 1011
Event: time 1605054206.497768, type 3 (EV_ABS), code 40 (ABS_MISC), value -1011
```
ABS_Z has a positive value corresponding with how much it's pressed down, while ABS_MISC has its inverse.

Right trigger (the variable one) sends:
```
Event: time 1605054246.638020, type 3 (EV_ABS), code 40 (ABS_MISC), value 104
Event: time 1605054246.638020, type 3 (EV_ABS), code 5 (ABS_RZ), value 104
```
ABS_RZ has a positive value corresponding with how much it's pressed down, with ABS_MISC matching.

> please also include the share button so we know the code of it
```
Event: time 1605053344.113912, type 4 (EV_MSC), code 4 (MSC_SCAN), value 90007
Event: time 1605053344.113912, type 1 (EV_KEY), code 314 (BTN_SELECT), value 1
Event: time 1605053344.113912, -------------- SYN_REPORT ------------
Event: time 1605053344.203915, type 4 (EV_MSC), code 4 (MSC_SCAN), value 90007
Event: time 1605053344.203915, type 1 (EV_KEY), code 314 (BTN_SELECT), value 0
Event: time 1605053344.203915, -------------- SYN_REPORT ------------
```

docs/descriptors/xb1sx.md
```diff
@@ -0,0 +1,167 @@
+# USB Descriptor for Xbox One S Wireless (Linux mode)
```
kakra, line 1: Please adjust the title.

lights0123:
This looks like it's using the select button code for the share button.

lights0123:
Forgot to mention that the thumb buttons work as well, with THUMBL and THUMBR matching the left and right button clicks respectively.

kakra:
> This looks like it's using the select button code for the share button.

Yeah, I was expecting that. Thus, we need to ignore it for now until we decided how to handle it. I can imagine two solutions: We'd map it to `BTN_TRIGGER_HAPPY` range (trigger is like joy, thus the name). But I'm unsure at which position: I'd like to have the four paddles of the XBE2 controller at `BTN_TRIGGER_HAPPY{1,2,3,4}`. So we'd need to map it to 5. But what if a future controller adds another gaming button?

So we should probably better move it to the keyboard range and use one of the `KEY_*` buttons for the share button. I'm not sure which, maybe there a screenshot keycode? At least that sound suitable. And after all: It's not a gaming button thus it should by no means have an `BTN_*` keycode.

lights0123:
It could be mapped to Print Screen, but that seems a bit weird. What if we just sent it to a FIFO instead? My [keyboard driver does this](https://github.com/ckb-next/ckb-next/wiki/CKB-Daemon-Manual#notifications), although in userspace. That way, you can just wait for new input there and handle it. I don't know if we want to deal with that complexity though for a single button (currently). It also complicates multiuser, as you can't just rely on the standard HID permission magic.

lights0123:
Actually, the behaviors that I noted seem to actually be correct. Just played a game, and everything including rumble seems to work—just the select button is weird.

kakra:
Now, let's deep dive...

> Yep, although I just got the controller by itself because a) all the other controllers were out of stock and b) might as well get the latest one—I don't actually own an Xbox, or plan to.

Ah, okay. We should refer to it as XBSX (Xbox Series X), lining up with XB1S and XBE2 using four-letter codes.

> > Also, please dump the device descriptor and document it in `docs/descriptors/`.
> 
> Done, although I don't know what I should call it.

Looks fine. If you don't mind, I'd rename it during merge if I feel it should have a different name.

> > Use evtest to test if each button maps correctly
> 
> The "change view" button (left button on face, I think also called Back/Select) simply had no output in `evtest`. In `btmon`, however, I got:
> 
> ```
> > ACL Data RX: Handle 3585 flags 0x02 dlen 23
>       ATT: Handle Value Notification (0x1b) len 18
>         Handle: 0x001e
>           Data: 857d887d887e06840000000000000000
> > ACL Data RX: Handle 3585 flags 0x02 dlen 23
>       ATT: Handle Value Notification (0x1b) len 18
>         Handle: 0x001e
>           Data: 857d887d887e06840000000000000400
> ```
> 
> The second one (with the `4`) had the button pressed.

This looks like it's using a different protocol, not the usual Bluetooth HID protocol. But at least Linux somehow seems to map it to HID so it ends up in xpadneo. Fine.

For the button bits: I wonder where button A is in the report. It is important to remap the keycodes in a way that, when ordered by keycode, we match the order A,B,X,Y,TL,TR,SELECT,START,THUMBL,THUMBR with no other keycodes generated inbetween. Otherwise, games and the kernel will mismatch the symbolic names with the buttons on the controller. This is why we cannot use `BTN_MODE` for the logo button because it comes before the thumb button and move their positional code, thus shifting select, start and thumb buttons by one position. This is what you're currently seeing with the share button. So we need to remap the share button out of this range. As I already wrote, and since it isn't a gaming button, I suggest moving it to a `KEY_*` range.

See also: https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h

> Select works. Y appears as West. B appears as East. X appears as North. A appears as South. Assuming that these are supposed to match with a compass, Y and X are swapped.

Yes, this is what the Linux gamepad specification says. But no, it doesn't match what games expect. Thus, for buttons please ignore the directional symbols, just use the named symbols.

> The DPad gives a positive value for down or right, and negative for up or left. Up and down are on the Y axis, left and right are on the X axis. This may mean that Y needs to be inverted.

No, I think that's okay: Up should be negative (this doesn't match with what the thumb axes do).

> The left joystick (ABS_X/ABS_Y) gives a positive value for up or right, and negative for down or left. Up and down are on the X axis, left and right are on the Y axis. This may mean that X and Y need to be swapped.
> The right joystick (ABS_RX/ABS_RY) has the same behavior.

This sounds strange... Let's fix some other stuff first and investigate the HID descriptor for this. Sounds like MS is trying to play tricks again with how Linux maps axes codes to actual symbolic axes. Or one of our HID fixups doesn't play nice with the new report format: We are actually swapping, moving, and shifting some codes around in the report fixup to convert it to a GIP compatible HID report.

> Left shoulder button sends BTN_TL, right shoulder button sends BTN_TR.

This is fine, that means there's no other `BTN_*` code interfering before these positions. OTOH, there are not many buttons that could do that except `BTN_C` and `BTN_Z` which are hopefully not assigned by the HID descriptor - and if they are, our report fixup should shift the bits to exclude those.

> Left trigger (the variable one) sends:
> 
> ```
> Event: time 1605054206.497768, type 3 (EV_ABS), code 2 (ABS_Z), value 1011
> Event: time 1605054206.497768, type 3 (EV_ABS), code 40 (ABS_MISC), value -1011
> ```
> 
> ABS_Z has a positive value corresponding with how much it's pressed down, while ABS_MISC has its inverse.

`ABS_MISC` is a synthesized axis we added due to a feature request to combine both triggers into a rudder axis. This can be used to assign both triggers to a single rolling axis in simulators. Since its positional code orders after the standard axis, this should create no compatibility problems.

> Right trigger (the variable one) sends:
> 
> ```
> Event: time 1605054246.638020, type 3 (EV_ABS), code 40 (ABS_MISC), value 104
> Event: time 1605054246.638020, type 3 (EV_ABS), code 5 (ABS_RZ), value 104
> ```
> 
> ABS_RZ has a positive value corresponding with how much it's pressed down, with ABS_MISC matching.

Matches what we expect. See above.

> > please also include the share button so we know the code of it
> 
> ```
> Event: time 1605053344.113912, type 4 (EV_MSC), code 4 (MSC_SCAN), value 90007
> Event: time 1605053344.113912, type 1 (EV_KEY), code 314 (BTN_SELECT), value 1
> Event: time 1605053344.113912, -------------- SYN_REPORT ------------
> Event: time 1605053344.203915, type 4 (EV_MSC), code 4 (MSC_SCAN), value 90007
> Event: time 1605053344.203915, type 1 (EV_KEY), code 314 (BTN_SELECT), value 0
> Event: time 1605053344.203915, -------------- SYN_REPORT ------------
> ```

Okay, so we cannot ignore it because it uses the code originally intended to be used by `BTN_SELECT`. So we need to use a report fixup to shift the reported bits around and order that bit after the standard button bits. This is were it gets a little more challenging now. You want to try?

You'll need to change `xpadneo_report_fixup()` to report 12 buttons for your controller. That function patches the descriptor you dumped to the documentation file: It matches some bytes in the report and patches those to report what we want it to report. You can validate the result using the online decoder. It's important to pad the count to a byte boundary probably, this is why we patch 15+1 bits to 11+5 bits, it makes 16 bits = 2 bytes. You'll have to use 12+4 probably.

It'll mess up the mappings or render button completely silent in `evtest` now. This is why you also need to patch the raw events from the device which happens in `xpadneo_raw_event()`: Adjust the code to shift the bit position 7 to 12: `00 40 ..` is at position 7 (counted backwards from the end of the word in `btmon`). It needs to become `08 00`. This change would make button positions 1-11 map correctly, matching what we define in `xpadneo_usage_maps`. If you look at `evtest` again now, it should report a new keycode for the share button. Now adjust that structure to map it something that is not a `BTN_*` code so we get it out of the range of gaming input. This is important because games need to see just 10 buttons, otherwise they may acts strange.

kakra:
> It could be mapped to Print Screen, but that seems a bit weird. What if we just sent it to a FIFO instead? My [keyboard driver does this](https://github.com/ckb-next/ckb-next/wiki/CKB-Daemon-Manual#notifications), although in userspace. That way, you can just wait for new input there and handle it. I don't know if we want to deal with that complexity though for a single button (currently). It also complicates multiuser, as you can't just rely on the standard HID permission magic.

We should probably move it to the range where the consumer control codes live, maybe `KEY_SENDFILE`? Sounds suitable...

kakra:
> It'll mess up the mappings or render button completely silent in `evtest` now. This is why you also need to patch the raw events from the device which happens in `xpadneo_raw_event()`: Adjust the code to shift the bit position 7 to 12: `00 40 ..` is at position 7 (counted backwards from the end of the word in `btmon`). It needs to become `08 00`. This change would make button positions 1-11 map correctly, matching what we define in `xpadneo_usage_maps`. If you look at `evtest` again now, it should report a new keycode for the share button. Now adjust that structure to map it something that is not a `BTN_*` code so we get it out of the range of gaming input. This is important because games need to see just 10 buttons, otherwise they may acts strange.

Actually, the current code may do something wrong with the share button already... But I'm guessing you'll be able to figure it out correctly.

docs/descriptors/xb1sx.md
```diff
@@ -0,0 +1,167 @@
...
+0x45, 0x00,        //   Physical Maximum (0)
+0x65, 0x00,        //   Unit (None)
+0x81, 0x03,        //   Input (Const,Var,Abs,No Wrap,Linear,Preferred State,No Null Position)
+0x05, 0x09,        //   Usage Page (Button)
+0x19, 0x01,        //   Usage Minimum (0x01)
+0x29, 0x0B,        //   Usage Maximum (0x0B)
+0x15, 0x00,        //   Logical Minimum (0)
+0x25, 0x01,        //   Logical Maximum (1)
+0x75, 0x01,        //   Report Size (1)
+0x95, 0x0B,        //   Report Count (11)
```
kakra, line 99: It reports 11 buttons here. This seems much more sane than the XBE2 report which says it has 15 buttons. It thus matches with the Windows mode of the older controllers, and the original non-wireless controllers. But it seems to put the share button in an unexpected position. You probably need to fix that to report 12 buttons instead, so we can patch the raw events to reports the 10 standard bits correctly, plus bit 11 for the logo button, and bit 12 for the share button.

docs/descriptors/xb1sx.md
```diff
@@ -0,0 +1,167 @@
...
+0x81, 0x02,        //   Input (Data,Var,Abs,No Wrap,Linear,Preferred State,No Null Position)
+0x15, 0x00,        //   Logical Minimum (0)
+0x25, 0x00,        //   Logical Maximum (0)
+0x75, 0x01,        //   Report Size (1)
+0x95, 0x05,        //   Report Count (5)
+0x81, 0x03,        //   Input (Const,Var,Abs,No Wrap,Linear,Preferred State,No Null Position)
+0x05, 0x0C,        //   Usage Page (Consumer)
```
kakra, line 106: This is probably the logo button. Thus, your descriptor won't report it in the 11 button bits above but in a separate report. That doesn't hurt, tho. We can still pretend 12 buttons above and live with a single blind bit. `xpadneo_usage_maps` will take care of this then and report the button. This usage page will probably report as `0x10085` or `0xC0223`. If not, add it to the struct. If it would be reported twice, add an ignore to the struct.

kakra (changes requested):
You should also add the controller to the README to report it working for other users, and document any missing features or special behavior.

kakra:
> Actually, the behaviors that I noted seem to actually be correct. Just played a game, and everything including rumble seems to work—just the select button is weird.

Yeah, games using `evdev` will probably work when they use symboling names to match the buttons. Modern SDL version can do that. But all games using `jsdev` or older SDL versions will use positional buttons for mapping, so we need to ensure that the first 10 positions map correctly, and in the best case we stay at 10 positions.

But I was thinking about the share button. It may actually be a game function, right? This is probably why MS decided to report 11 buttons instead of 10, so games will see a button 11 and use that. In that case, it should be mapped to `BTN_TRIGGER_HAPPY#' - whatever that number is because XBE2 has the four paddle buttons. We probably need to figure that out later when games start to support the additional buttons. Until then, let's move it to a `KEY_*` and leave a comment in the code about that.

kakra:
After `evtest` works correctly, test if `jstest` works correctly. It should report no bogus buttons or axes, and match positions correctly with one caveat: You would see axes positions 1-6 as `LX,LY,LT,RX,RY,RT` instead of `LX,XY,RX,RY,LT,RT` (which is what jsdev games expect). But that's to be fixed with `jscal`. We cannot reorder jsdev from within xpadneo in a compatible way.

kakra:
> We should probably move it to the range where the consumer control codes live, maybe `KEY_SENDFILE`? Sounds suitable...

`KEY_RECORD` could make a good code, too. I think the share button is intended to work this way: Short press = record screenshot, long press = record video. So "record" makes most sense to me.

lights0123:
Here's SDL's change: https://hg.libsdl.org/SDL/rev/38671e8e9a31#l6.34

lights0123:
> we match the order A,B,X,Y,TL,TR,SELECT,START,THUMBL,THUMBR

I can confirm that is matched by https://gamepad-tester.com/.

>  >   The left joystick (ABS_X/ABS_Y) gives a positive value for up or right, and negative for down or left. Up and down are on the X axis, left and right are on the Y axis. This may mean that X and Y need to be swapped.
> >    The right joystick (ABS_RX/ABS_RY) has the same behavior.
>
> This sounds strange... Let's fix some other stuff first and investigate the HID descriptor for this. Sounds like MS is trying to play tricks again with how Linux maps axes codes to actual symbolic axes. Or one of our HID fixups doesn't play nice with the new report format: We are actually swapping, moving, and shifting some codes around in the report fixup to convert it to a GIP compatible HID report.

The left joystick at least worked properly in Rocket League under Wine, so that may also be working as expected.

--

I'm working on figuring out what modifications I need to make right now.

kakra:
> Here's SDL's change: https://hg.libsdl.org/SDL/rev/38671e8e9a31#l6.34

Interesting, so they are putting it in front of the paddle buttons but conditionally? That sounds like a strange decision because it maps the paddles to position 11-14 for controllers without a share button and 12-15 for controllers with a share button (or +1 if we count the guide button).

kakra:
> > we match the order A,B,X,Y,TL,TR,SELECT,START,THUMBL,THUMBR
> 
> I can confirm that is matched by https://gamepad-tester.com/.

At least in Chrome, Gamepad Tester is its own story... It makes some wrong assumptions: It reads a mapping from evdev but then uses jsdev to read the actual axes and buttons despite jsdev having an actual API to get the symbolic names. Since that results in some mis-mappings, they've put code in to correct the wrong mappings when some device VID/PID matches - so they are adding an additional layer of hardware abstraction. This doesn't work with xpadneo correctly: It confuses the axes mapping, and you cannot correctly fix it with jscal now because they combine evdev+jsdev with a custom hardware database.

kakra:
> This doesn't work with xpadneo correctly: It confuses the axes mapping, and you cannot correctly fix it with jscal now because they combine evdev+jsdev with a custom hardware database.

Either it works now, or the XBE2 controller doesn't have wrong assumptions in the Chrome Gamepad API.

lights0123:
> You'll need to change `xpadneo_report_fixup()` to report 12 buttons for your controller..

How should I gate that, like you do here?

https://github.com/atar-axis/xpadneo/blob/63fceb1580491d306f42b10d9821a9160e58ad1d/hid-xpadneo/src/hid-xpadneo.c#L841-L846

 Should I check if there's 11 buttons, and then change it to 12? Should I look at the product ID?

kakra:
> How should I gate that, like you do here?

Yeah, just look if there's 11 buttons and change to 12. That's the correct way to do it. Maybe add a quirk flag for the share button so you can pull it from xdata in the other functions.

I just realized in one of your first comment that "change page" is not "share" - so some of my comments may confuse the bit positions. But I think you get the picture.

lights0123:
wait, I just realized that the descriptor message I posted must include that fixup already, right? `dmesg` includes "fixing up button mapping", but the dump I got shows 11 buttons, so that must mean that it already includes its changes.

https://github.com/atar-axis/xpadneo/blob/63fceb1580491d306f42b10d9821a9160e58ad1d/hid-xpadneo/src/hid-xpadneo.c#L841-L854

That means that I can't just rely on a 11→12 change as it appears identical, so I'll have to switch either based on some descriptor data if it's present, or product ID. Also, I should probably re-dump the descriptors. Is there an easy way to do that? Can I unload the kernel module and get it somehow, or should I comment out that block?

lights0123:
Unless you're OK with other controllers *also* having 12 buttons, which I assume you don't want.

kakra:
> That means that I can't just rely on a 11→12 change as it appears identical, so I'll have to switch either based on some descriptor data if it's present, or product ID. Also, I should probably re-dump the descriptors. Is there an easy way to do that? Can I unload the kernel module and get it somehow, or should I comment out that block?

If your kernel has debug mode enabled, you MAY get the original rdesc from `/sys/kernel/debug/hid`, but please double-check. Otherwise you'll have to blacklist xpadneo temporarily to prevent it from loading.

Only rely on PID if no other option is available because MS tends to change the PID with firmware updates sometimes without updating the rdesc, and at least at one point they changed the rdesc without updating the PID. So the only reliable way is checking if the expected bytes at the expected positions are actually present.

Just put your patch above the "15->11" and it should work, maybe put a comment to hint importance of order. Positions 145 and 163 won't match my patch then.

kakra:
> Unless you're OK with other controllers _also_ having 12 buttons, which I assume you don't want.

We can add a quirk flag to tell such cases apart at any time later... In the "rdesc>=164" case we should use if/else to tell both models apart, and either patch 10->11 or 10->12 buttons. Depending on the original unpatched descriptor, you may need to set the linux buttons quirk flag then. Maybe also adjust the documentation to have the unpatched descriptor instead.

lights0123:
Just published the *actual* descriptor.

docs/descriptors/xb1sx.md
```diff
@@ -92,16 +92,16 @@ Parsed descriptor:
 0x81, 0x03,        //   Input (Const,Var,Abs,No Wrap,Linear,Preferred State,No Null Position)
 0x05, 0x09,        //   Usage Page (Button)
 0x19, 0x01,        //   Usage Minimum (0x01)
-0x29, 0x0B,        //   Usage Maximum (0x0B)
+0x29, 0x0F,        //   Usage Maximum (0x0F)
 0x15, 0x00,        //   Logical Minimum (0)
 0x25, 0x01,        //   Logical Maximum (1)
 0x75, 0x01,        //   Report Size (1)
-0x95, 0x0B,        //   Report Count (11)
```
kakra: Ah okay, so the original rdesc has 15 buttons and MS just uses one of the 15 bits to report the button.

hid-xpadneo/src/hid-xpadneo.c
kakra: You may need to add the PID to the "pretend XB1S mode" switch in `xpadneo_input_configured()` if SDL applies wrong mappings after your changes.

kakra:
> or should I comment out that block?

Of course that should work, too.

hid-xpadneo/src/hid-xpadneo.c
kakra: Or pretend a different firmware version...

lights0123:
5808dd3 ignores the share button, but uses the physical select button. Only switching based on PID is implemented so far.

I'm out for the night. I'm not totally sure how to report the share button status—do I just set it as `data[15] & BIT(12-8)`?

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -847,9 +849,14 @@ static u8 *xpadneo_report_fixup(struct hid_device *hdev, u8 *rdesc, unsigned int
 			struct xpadneo_devdata *xdata = hid_get_drvdata(hdev);
 			hid_notice(hdev, "fixing up button mapping\n");
 			xdata->quirks |= XPADNEO_QUIRK_LINUX_BUTTONS;
-			rdesc[145] = 0x0B;	/* 15 buttons -> 11 buttons */
-			rdesc[153] = 0x0B;	/* 15 bits -> 11 bits buttons */
-			rdesc[163] = 0x05;	/* 1 bit -> 5 bits constants */
+			u8 button_count = hdev->product == 0x0B13 ? 12 : 11;
```
kakra, line 852: Just make it 12 bits without looking at the PID.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -899,7 +906,12 @@ static int xpadneo_raw_event(struct hid_device *hdev, struct hid_report *report,
 		bits |= (data[14] & (BIT(0) | BIT(1))) >> 0;	/* A, B */
 		bits |= (data[14] & (BIT(3) | BIT(4))) >> 1;	/* X, Y */
 		bits |= (data[14] & (BIT(6) | BIT(7))) >> 2;	/* LB, RB */
-		bits |= (data[16] & BIT(0)) << 6;	/* Back */
+		if (xdata->quirks & XPADNEO_QUIRK_SHARE_BUTTON) {
```
kakra, line 909: Do we really need a quirk flag for that? What do do the back and share buttons look like in `btmon`?

kakra (commented):
Could you decode all the button bits in the `btmon` reports and document with the report descriptor in `docs/`?

hid-xpadneo/src/hid-xpadneo.c
kakra: IOW, for my controller, the bit positions of A,B,X,Y,LB,RB,Back.Menu,LS,RS,Xbox are 0,1,3,4,6,7,16,11,13,14,12. Does this match your controller? Do order that correctly by symbolic order, I'm shifting the bits to 0..10, thus share should go to 11. (counted from 0 as in the code)

kakra: Don't test with games yet at this stage, nor SDL. `evtest` and `jstest` have to work correctly first.

lights0123: We need a quirk flag because the share button is `data[16] & BIT(0)`, the same as the select button on other models. The actual select button is at `data[15] & BIT(2)`, which as far as I can tell is unused by other models.

kakra: > 7,16,11

And yes, the controller has a button bit in a position that's not covered by the HID descriptor... But only with some firmwares...

kakra: > We need a quirk flag because the share button is `data[16] & BIT(0)`, the same as the select button on other models. The actual select button is at `data[15] & BIT(2)`, which as far as I can tell is unused by other models.

So MS plays funny games again... `data[16] & BIT(0)` has the back button on SOME firmware versions, on others it's sent in a separate report (opcode 2 instead of 1 I believe) as a consumer control code (AC Back). So, yes: In that case a quirk flag is perfectly suitable.

So in the HID descriptor patcher just make it 12 buttons than - no matter which controller we have. Since we cannot tell the models apart here and it feels wrong looking at the PID at this point, you should move the quirk flag to the `xpadneo_quirks` struct and put an OUI matcher: You model has a different OUI than other models.

Then in the raw event patcher, do not try to play too many tricks: You'll either get the share button at `data[15] & BIT(0)` and move it to bit 11(+1), or you'll get the back button and move it to bit 10(+1). (+1 unless counting from 0). It looks like your back button either comes in report opcode 2, or is at a different position. The code should be clear and easy to understand, I'm currently having problems grasping the purpose of
```c
    bits |= (data[15] & BIT(2)) ? BIT(6) : 0; /* Back */
} else {
    bits |= (data[16] & BIT(0)) << 6; /* Back */
```

Maybe you wanted `bits |= (data[15] & BIT(2)) << 4;` which moves bit 2 of byte 15 to bit 6 of `u16 bits`? Because your back button is at bit 2 byte 15?

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -899,7 +906,12 @@ static int xpadneo_raw_event(struct hid_device *hdev, struct hid_report *report,
...
+			/* correct button mapping of Xbox controllers with a share button */
+			bits |= (data[15] & BIT(2)) ? BIT(6) : 0; /* Back */
```
kakra, line 911: Looks complicated... Please try to avoid the ternary operator to reflect how the other lines do it. It took me some minutes to realize what this says. ;-) It's probably more easy to read (put bit 6 if bit 2 is set) but the other lines shift bits instead of explicitly setting them.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -899,7 +906,12 @@ static int xpadneo_raw_event(struct hid_device *hdev, struct hid_report *report,
 		bits |= (data[14] & (BIT(0) | BIT(1))) >> 0;	/* A, B */
 		bits |= (data[14] & (BIT(3) | BIT(4))) >> 1;	/* X, Y */
 		bits |= (data[14] & (BIT(6) | BIT(7))) >> 2;	/* LB, RB */
-		bits |= (data[16] & BIT(0)) << 6;	/* Back */
+		if (xdata->quirks & XPADNEO_QUIRK_SHARE_BUTTON) {
+			/* correct button mapping of Xbox controllers with a share button */
```
kakra, line 910: Remove this comment, then we can also skip using the curly braces.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -899,7 +906,12 @@ static int xpadneo_raw_event(struct hid_device *hdev, struct hid_report *report,
...
+			bits |= (data[15] & BIT(2)) ? BIT(6) : 0; /* Back */
+		} else {
+			bits |= (data[16] & BIT(0)) << 6; /* Back */
+		}
 		bits |= (data[15] & BIT(3)) << 4;	/* Menu */
 		bits |= (data[15] & BIT(5)) << 3;	/* LS */
 		bits |= (data[15] & BIT(6)) << 3;	/* RS */
```
kakra, line 917: The order of the above lines reflects the positional orders of buttons games expect. So the share button should probably somewhere after here, even if that means we need another `if quirk-sharebutton`.

kakra (changes requested):

hid-xpadneo/src/hid-xpadneo.c
kakra: Or it doesn't quite so... Menu should move down. Let's ignore that for now and do better from now on. ;-)

hid-xpadneo/src/hid-xpadneo.c
lights0123: I found it the opposite way when reading it: the new position is dependent on the old one, and I had to do math in my head to figure out where that was. This was just my first get-it-working pass, so I didn't bother to calculate the bitshift. I'll do that though tomorrow.

kakra: Yeah, I figured that out eventually. I came from the other side when I developed this piece of code because we had a bit mask of
```
...11011011
...RLZYXCBA
```
and I've just gone: "okay, take these two, skip 1, take the next 2, skip 1 - ah okay, that shift 1, then shift 2..."

Also, I think the resulting assembler code is better optimized without using ternary operators. So I decided to continue that way even if that meant to do some math in the head. :-)

Sleep well...

cow-killer:
So, how is the progress coming along? Just got myself the Series X pad yesterday and wondering if these commits will incorporate Bluetooth functionality.

kakra:
@cow-killer xpadneo does work ONLY with Bluetooth - so yes, it will support this as long as your Bluetooth dongle supports the BLE profile (Bluetooth Low Energy). I think the XBXS controller is BLE-only (or the proprietary 5 GHz dongle but that's not covered by xpadneo currently).

lights0123:
@cow-killer if you're impatient, the code in this PR does work—the only thing left really is a bit of code cleanup and figuring out what to do with the share button.

kakra:
@cow-killer This PR is not even 24 hours old - give us a little air to breathe :-D

hid-xpadneo/src/hid-xpadneo.c
lights0123: To be fair, ternaries [generate](https://godbolt.org/z/Pjc1no) fewer instructions without optimization and identical instructions with even `-O1` :smile:.

lights0123:
I've added
```c
if (xdata->quirks & XPADNEO_QUIRK_SHARE_BUTTON)
	bits |= (data[16] & BIT(0)) << 11; /* Share */
```
how do I actually get that to show up in e.g. `evtest` now?

kakra:
> how do I actually get that to show up in e.g. `evtest` now?

It should generate a code in `evtest` already but the wrong one... Did you test?

hid-xpadneo/src/hid-xpadneo.c
kakra: Oh, I didn't know... Then maybe it's time to rewrite all the lines. It's up to you if you want to do it or if I do it.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -202,6 +202,7 @@ static const struct quirk xpadneo_quirks[] = {
 	DEVICE_OUI_QUIRK("E4:17:D8",
 			 XPADNEO_QUIRK_NO_PULSE | XPADNEO_QUIRK_NO_TRIGGER_RUMBLE |
 			 XPADNEO_QUIRK_NO_MOTOR_MASK | XPADNEO_QUIRK_NINTENDO),
+	DEVICE_OUI_QUIRK("44:16:22", XPADNEO_QUIRK_SHARE_BUTTON),
```
kakra, line 205: Does the gamepad have rumble motors in the triggers? If not, we may want to set the no-trigger-rumble quirk, too.

kakra (changes requested):
Please add the additional known PIDs to the comments in `xpadneo_input_configured()`.

lights0123:
I did. With some `printk` debugging, I can see that `bits` is set to 2048, but no output from `evtest`.

hid-xpadneo/src/hid-xpadneo.c
lights0123: Is there a way to test if it does?

lights0123:
In `xpadneo_input_configured`, what should I have it change its PID to?

hid-xpadneo/src/hid-xpadneo.c
kakra: Cool site, btw... When I researched that, people always insisted that ternaries generate additional code overhead that cannot be properly optimized, and that it generates jumps the hurts the branch prediction performance... Well, looks like things may actually be different. Tho, your example is optimized for an immediate return and could also be inlined by the compiler. I'm not sure what the cost of a failed branch prediction is. Also, that example evaluates statically: In the O1 case, both are identical.

kakra:
> In `xpadneo_input_configured`, what should I have it change its PID to?

This would be needed if SDL mappings are wrong. The use the `0x0B05` case, I'd say. Or pretend a different firmware version: SDL matches devices by firmware version, too, not just VID/PID. Chrome matches by PID/VID only.

hid-xpadneo/src/hid-xpadneo.c
kakra: You'd either need a screwdriver, or you use the hidraw rumble tester from the misc folder.

kakra:
> I did. With some `printk` debugging, I can see that `bits` is set to 2048, but no output from `evtest`.

This is strange... Maybe this is a result from struct `xpadneo_usage_maps` setting an event code to ignore?

hid-xpadneo/src/hid-xpadneo.c
lights0123: How do I get a hidraw node? With this module loaded or unloaded, I don't see a new /dev/hidraw* node.

lights0123:
Oh yeah, I just added it in `xpadneo_usage_maps` to `KEY_RECORD` and it works now.

hid-xpadneo/src/hid-xpadneo.c
lights0123: Ah, it's `hidraw0`. Yes, it does have trigger rumbles.

kakra:
> Oh yeah, I just added it in `xpadneo_usage_maps` to `KEY_RECORD` and it works now.

\o/ yay! Okay, time for testing now. First things first:

Does `evtest` show correct symbols for each and every key and axis?

If yes, does `jstest` actually show the expected positional order of buttons (that is, 10 standard buttons followed by optional extra buttons), and does it have the first 6 axis in standard order (X,Y,Z,Rx,Ry,Rz aka left stick x,y, left trigger, right stick x,y, right trigger)?

If yes, test SDL apps, Wine, maybe Gamepad Tester in the browser.

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -239,6 +239,7 @@ static const struct usage_map xpadneo_usage_maps[] = {
 
 	/* fixup the Xbox logo button */
 	USAGE_MAP(0x9000B, MAP_STATIC, EV_KEY, BTN_XBOX),	/* Xbox */
+	USAGE_MAP(0x9000C, MAP_STATIC, EV_KEY, KEY_RECORD),	/* Share */
```
kakra, line 242: Comment above should mention the share button, maybe insert an additional line.

kakra (changes requested):
Looks fine at a first glance, now run it through `indent -v` relative to the project root. This should fix spaces and line lengths to kernel standards. It may act strange in one place in the code.

lights0123:
> Does `evtest` show correct symbols for each and every key and axis?

yes
> If yes, does jstest actually show the expected positional order of buttons (that is, 10 standard buttons followed by optional extra buttons), and does it have the first 6 axis in standard order (X,Y,Z,Rx,Ry,Rz aka left stick x,y, left trigger, right stick x,y, right trigger)?

yes
> test SDL apps

Downloading supertuxkart right now, for some reason Arch chose to download from a mirror at less than 3Mbps when I have a gigabit connection. If you have an app that you've tested with, I can do that too.
> Wine

Rocket League works
> Gamepad tester

works in Firefox and Chrome

kakra:
Very cool. Thanks for the hard work. I'll mention you in the NEWS file of the next release by full name if that's okay. Also, I'm going to test if my devices still work this evening before merging.

If you want to sqash some commits, or rewrite history, feel free to do so. Otherwise I'm going to squash some of your commits where I feel appropriate.

lights0123:
supertuxkart works, it just doesn't have the button names (says e.g. "Gamepad button 5")

hid-xpadneo/src/hid-xpadneo.c
```diff
@@ -1222,18 +1224,18 @@ static void xpadneo_remove(struct hid_device *hdev)
 
 static const struct hid_device_id xpadneo_devices[] = {
 	/* XBOX ONE S / X */
-	{HID_BLUETOOTH_DEVICE(USB_VENDOR_ID_MICROSOFT, 0x02FD)},
```
kakra, line 1215: This is where my indent version acts different to yours. I don't know why, but have seen it before.

lights0123: Version differences?

kakra: Yeah, maybe... We should probably drop these changes so it doesn't play ping-pong.

kakra:
> If you have an app that you've tested with, I can do that too.

I'm always seeing problems with AC:Unity and AC:BlackFlag... These seem to see the triggers as a joystick axis, I don't know why but you'd need to press them half-way to stop erratic movement and turns. But that seems to be a Wine thing, it worked properly with some older versions.

Also, `jstest-gtk` will probably display axes wrongly. This can be fixed with jscal by putting a different order. We apparently cannot do that from the driver, and also cannot ship it as a udev rule because it breaks Gamepad Tester.

lights0123:
Side note, what do you think would be the easiest way to hook the share button up to do *something*? GNOME's hotkey feature doesn't do anything when I press it. Is there some way I can write a script that measures how long it's pressed down and does some action based on that? I could read `/dev/input/event*`, but that requires root.

kakra (approved):
The commits are sitting in my test queue for now. I'll probably close this PR by merging from the queue branch, so you'll see this closes with unmerged commits. I squashed most of your commits and split out one hunk to put it in front of the other commits. Hope you're fine with that: #254

kakra:
First tests show it's properly working in `evtest` and `jstest` for my controller. While testing, I found a funny bug: If you depress both thumb stick buttons while the Steam client is running, you're able to move the mouse pointer and do clicks. I've reported that to Valve, I'm not sure why the desktop controller profile of Steam input kicks in when doing that.

Apparently, I don't have too much time this evening, so I'm doing game tests later.

cow-killer:
@kakra Getting this when trying to run `install.sh` after cloning your repository:
`sed: can't read ./lib/../VERSION: No such file or directory`
Any chance you could help me out here?

lights0123:
@cow-killer are you sure that you *cloned* it (rather than using GitHub's download from master), and `git` recognizes it as a repository?

kakra:
Please file a new bug report for this.

lights0123:
@kakra that's a feature.
Screenshot from 2020-11-11 18-05-13

kakra:
@lights0123 yeah but I disabled that for the Xbox controller, and it only happens when I hold down both thumb stick buttons.

lights0123:
Well... spoke too soon. I just opened up Wine's own Gamepad manager, and it sees the share button as button 1. Apparently [this is fixed in Proton](https://www.reddit.com/r/linux_gaming/comments/hjjevn/wine_control_panel_controller_configuration_tech/fwnb7bt), so I may just go with the creating-an-SDL-configuration method or seeing if I can get the Epic Games Store running under Proton with Lutris.

kakra:
> Well... spoke too soon

@lights0123 I merged your changes. If anything is missing or needs fixing, we should open a new report for that.

lights0123:
That actually seems to be a non-issue: apparently this game is known to have gamepad issues even on Windows natively.

PR #258, opened 2020-11-15
Document BLE requirements

kakra:


atar-axis:
I havent tried this on any adapter without BLE support, but the following might work:

`sudo btmgmt le on`

or: `sudo btmgmt commands | grep -i "low energy"`
or: `sudo btmgmt info | grep -E "(supported settings)(.*) le "`

But let's verify that by someone who has an old adapter without BLE

Issue #268, opened 2021-01-10
xbox series x controller 045E:0B13 wrongly mapped


TTTPOB:
## Version of xpadneo

2020.11.22

## Severity / Impact

* [x] I've read the docs and the bug reporting instructions
* [ ] It does not work at all
* [ ] It used to work in a previous version
* [ ] It mostly works but sometimes it doesn't
* [ ] I found a work-around
* [ ] I probably didn't figure it all out but it's too early to give up
* [ ] I don't know how to ... 
* [ ] It's too complicated
* [ ] Fantastic work but ... 
* [ ] I can code and I want to help

## Describe the bug

xbox series x controller 0005:045E:0B13.0005 works with usb cable but wrongly mapped with bluetooth

it reported it has 9 axis, and axis 9 will be affected by axis 3(lt) and axis 6(rt). back key was not responsible. share key mapped to key 7, left stick and right stick works normally. dpad seems normally too
I cannot see how other keys which I didnt mentioned work.

## System information

```console
# uname -a
Linux laptop 5.10.5-arch1-1 #1 SMP PREEMPT Thu, 07 Jan 2021 09:50:43 +0000 x86_64 GNU/Linux
```

```console
# xxd -c20 -g1 /sys/module/hid_xpadneo/drivers/hid:xpadneo/0005:045E:*/report_descriptor | tee >(cksum)
00000000: 05 01 09 05 a1 01 85 01 09 01 a1 00 09 30 09 31 15 00 27 ff  .............0.1..'.
00000014: ff 00 00 95 02 75 10 81 02 c0 09 01 a1 00 09 33 09 34 15 00  .....u.........3.4..
00000028: 27 ff ff 00 00 95 02 75 10 81 02 c0 05 01 09 32 15 00 26 ff  '......u.......2..&.
0000003c: 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01 81 03 05 01 09  ...u.....%.u........
00000050: 35 15 00 26 ff 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01  5..&....u.....%.u...
00000064: 81 03 05 01 09 39 15 01 25 08 35 00 46 3b 01 66 14 00 75 04  .....9..%.5.F;.f..u.
00000078: 95 01 81 42 75 04 95 01 15 00 25 00 35 00 45 00 65 00 81 03  ...Bu.....%.5.E.e...
[...]

PR #269, opened 2021-01-10
hid-xpadneo: Move share button quirk to `xpadneo_devices` database

kakra:
It looks like XBXS controllers are shipped using two different OUIs, one of those conflicts with the XBE2 controller. Detect the device quirk by PID instead.

Fixes: https://github.com/atar-axis/xpadneo/issues/268


kakra:
@lights0123 Could you verify that this commit doesn't change behavior of your controller?

lights0123:
@kakra can confirm that this works fine

PR #283, opened 2021-03-28
Allow the driver to operate in Xpad360 emulation mode

kakra:
Todo:

* [ ] Maybe find a way to implement this as a quirk flag
* [ ] Maybe make it switchable at runtime (can currently only be set at module load time)
* [ ] Investigate integration with https://github.com/atar-axis/xpadneo/issues/286


kakra:
xpadneo now has options in place that should work around any left issues with this. The only thing left is using 8-bit values for the triggers, this should be fixed in user-space. Future customization option may add additional improvements. Closing.

PR #288, opened 2021-04-29
Don't disable ERTM if kernel 5.12 or later

Mikaka27:
Removed disabling of ERTM only from dkms.post_install, I left dkms.post_remove alone. Reasoning for this was, that user could have installed xpadneo on older kernel, updated their kernel to 5.12 or later, and did an "sudo ./update.sh", which has to remove old ERTM override.

### DIFF
--- a/hid-xpadneo/dkms.post_install
+++ b/hid-xpadneo/dkms.post_install
@@ -3,7 +3,9 @@
 ERTM_OVERRIDE="/etc/modprobe.d/99-xpadneo-bluetooth.conf"
 DISABLE_ERTM="/sys/module/bluetooth/parameters/disable_ertm"
 
-if [ "$(readlink "${ERTM_OVERRIDE}" 2>/dev/null)" = "/dev/null" ]; then
+if [ `uname -r | awk -F. '{ printf "%03d%03d",$1,$2 }'` -ge 005012 ]; then
+	echo "Not disabling ERTM, kernel version doesn't require it..."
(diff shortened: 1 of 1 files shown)


hid-xpadneo/dkms.post_install
```diff
@@ -3,7 +3,12 @@
 ERTM_OVERRIDE="/etc/modprobe.d/99-xpadneo-bluetooth.conf"
 DISABLE_ERTM="/sys/module/bluetooth/parameters/disable_ertm"
 
-if [ "$(readlink "${ERTM_OVERRIDE}" 2>/dev/null)" = "/dev/null" ]; then
+KERNEL_MAJOR=`uname -r | awk -F'.' '{print $1}'`
+KERNEL_MINOR=`uname -r | awk -F'.' '{print $2}'`
+
+if [ "$KERNEL_MAJOR" -ge 6 ] || ([ "$KERNEL_MAJOR" -ge 5 ] && [ "$KERNEL_MINOR" -ge 12 ]); then
```
kakra, line 9: I'd probably use something like `uname -r | awk -F. '{ printf "%03d%03d",$1,$2 }'` and compare against `005012`.

kakra (changes requested):
My suggestion makes it easier to compare against multiple versions, in fact, if the L2CAP patch would be back-ported, you could just check a range `5010xxx..5011000` (would require to add the third version component).

Mikaka27:
I've updated the pull request according to your comments. It's better that way.

kakra (approved):
LGTM

kakra:
Can you squash both commits?

Mikaka27:
Just did :)

Issue #291, opened 2021-05-19
Incorrect buttons mapping on XBox Series X controller


anton-gorbikov:
## Version of xpadneo
0.9

## Severity / Impact
Medium

* [x] I've read the docs and the bug reporting instructions
* [x] It does not work at all
* [ ] It used to work in a previous version
* [ ] It mostly works but sometimes it doesn't
* [ ] I found a work-around
* [x] I probably didn't figure it all out but it's too early to give up
* [ ] I don't know how to ... 
* [ ] It's too complicated
* [x] Fantastic work but ... 
* [ ] I can code and I want to help

## Describe the bug
Some buttons have incorrect mappings, some don't work at all (probably unmapped):
- LB -> Y
- RB -> none
- share -> LB
- menu -> RB
- back -> none
- Y - X
- X -> none
All other buttons are working fine.

## Steps to Reproduce
1. Connect the wireless XBox Series X controller to Raspberry Pi 4. 
2. Run Steam Link
3. Run any game or controller test

## Expected behavior
Buttons should be mapped properly

## System information

```console
# uname -a
Linux raspberrypi 5.10.17-v7l+ #1414 SMP Fri Apr 30 13:20:47 BST 2021 armv7l GNU/Linux
```

```console
# xxd -c20 -g1 /sys/module/hid_xpadneo/drivers/hid:xpadneo/0005:045E:*/report_descriptor | tee >(cksum)
00000000: 05 01 09 05 a1 01 85 01 09 01 a1 00 09 30 09 31 15 00 27 ff  .............0.1..'.
00000014: ff 00 00 95 02 75 10 81 02 c0 09 01 a1 00 09 33 09 34 15 00  .....u.........3.4..
00000028: 27 ff ff 00 00 95 02 75 10 81 02 c0 05 01 09 32 15 00 26 ff  '......u.......2..&.
0000003c: 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01 81 03 05 01 09  ...u.....%.u........
00000050: 35 15 00 26 ff 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01  5..&....u.....%.u...
00000064: 81 03 05 01 09 39 15 01 25 08 35 00 46 3b 01 66 14 00 75 04  .....9..%.5.F;.f..u.
00000078: 95 01 81 42 75 04 95 01 15 00 25 00 35 00 45 00 65 00 81 03  ...Bu.....%.5.E.e...
[...]

PR #292, opened 2021-05-22
Fix issues with Steam Link mapping and Consumer Control detection

kakra:


crwxrws:
@kakra I can confirm that an extra "keyboard" input appears, and both the Xbox button and the Share button send events.

The Xbox button (KEY_MODE) sends a pair of key-down and key-up events exactly when the Xbox button is released. It does nothing until I release the button, i.e. it doesn't send a key-down event when I first press it.

The Share button (KEY_RECORD) sends a pair of key-down and key-up events exactly when it is pressed. It does nothing when released.

kakra:
> The Xbox button (KEY_MODE) sends a pair of key-down and key-up events exactly when the Xbox button is released. It does nothing until I release the button, i.e. it doesn't send a key-down event when I first press it.

Yes, this is expected: I implemented it this way to make it act as a shift key. If you press and hold it, then press A, B, X, or Y, it will switch to a different mapping profile (you can watch this event with `dmesg -w`). In that case, the Xbox button will not send an event when released. The different mappings itself are not implemented yet, tho. It's part of the profile emulation I'm planning to get feature parity with the Xbox Elite 2 controller which has a native profile switcher.

kakra:
> > The Xbox button (KEY_MODE) sends a pair of key-down and key-up events exactly when the Xbox button is released. It does nothing until I release the button, i.e. it doesn't send a key-down event when I first press it.
> 
> Yes, this is expected: I implemented it this way to make it act as a shift key. If you press and hold it, then press A, B, X, or Y, it will switch to a different mapping profile (you can watch this event with `dmesg -w`). In that case, the Xbox button will not send an event when released. The different mappings itself are not implemented yet, tho. It's part of the profile emulation I'm planning to get feature parity with the Xbox Elite 2 controller which has a native profile switcher.

> The Share button (KEY_RECORD) sends a pair of key-down and key-up events exactly when it is pressed. It does nothing when released.

This is a solution to a technical problem because your model reports the Share key on the gamepad device but we need it on the consumer device for compatibility among all the models. The Xbox Elite 2 and Xbox One S controller both have a separate gamepad and consumer sub device, your model doesn't, so it's emulated/synthesized. It's too difficult to maintain button state across both a real and a synthetic device but I may look into that again later.

crwxrws:
Amazing! Thank you very much for putting so much effort into this project. I'm already very happy with its current state. Even though personally I have no need for these two buttons, it's nice to see that they may become more useful. You can ping me if you need more input or testing, I will help if I have time.

AlexandreLaborde:
Hi @kakra its me from #291 😀
My controller's firmware just updated to 5.7.2688.0 and it causing issues with some Bluetooth devices. Let me know if you want/need my help to troubleshoot things on a Pi4.
Thanks for the great work!

kakra:
> Let me know if you want/need my help to troubleshoot things on a Pi4.

I don't think this belongs here... There are a number of Pi4-related issues about Bluetooth connection problems already, you should comment there. Also, this is most likely not something that xpadneo can solve, you could ask in our Discord community, or check the bluez project. I should probably make getting a Pi4 one of the next goals in ko-fi so I can better support issues with that hardware.

kakra:
Did you test on a RPi so we can get that checkbox done in the first post?

Mskeffington:
Oh, yes, sorry, I missed your whole checkbox system.  All my testing has been on RPi 3. My RPi 4 is giving me a weird bluez failure so I cant pair the controller correctly.

I have not tested this on x86 because my series X controller doesn't pair correctly.

PR #327, opened 2021-11-23
Update install.sh

jcfrosty:
New DKMS causing me grief in Manjaro attempting to compile/install module v0.9.1 . Read up on the manual for a quick fix and update here, only to find there was sort of a fix already here. Figured I could help clean it up some.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,10 +11,13 @@ if [[ -z "${INSTALLED[*]}" ]]; then
     echo "* creating dkms.conf"
     sed 's/"@DO_NOT_CHANGE@"/"'"${VERSION}"'"/g' <hid-xpadneo/dkms.conf.in >hid-xpadneo/dkms.conf
 
+    echo "* installing sources (using DKMS)"
     # TODO: Works around https://github.com/dell/dkms/issues/177 for DKMS 3
-    echo "* adding hid-xpadneo-${VERSION} folder to /usr/src"
-    mkdir -p "/usr/src/hid-xpadneo-${VERSION}"
-    cp --recursive "${V[@]}" hid-xpadneo/. "/usr/src/hid-xpadneo-${VERSION}/."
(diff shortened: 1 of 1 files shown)


kakra:
There's already a fix in the master branch which also passes the test. Your fix is missing at least one other change. Please wait until I backported the patches, I want to get at least one more fix into 0.9.x for the new Xbox controller models but it looks like that may take a while.

Here's the proposal for v0.9.2: https://github.com/atar-axis/xpadneo/pull/325 - as such closing this PR as a fixed is already existing. You may want to try if it works for you, our CI currently doesn't cover the new DKMS version (which, btw, seems to be buggy, an updated DKMS may also fix your problem).

install.sh
```diff
@@ -11,10 +11,13 @@ if [[ -z "${INSTALLED[*]}" ]]; then
     echo "* creating dkms.conf"
     sed 's/"@DO_NOT_CHANGE@"/"'"${VERSION}"'"/g' <hid-xpadneo/dkms.conf.in >hid-xpadneo/dkms.conf
 
+    echo "* installing sources (using DKMS)"
     # TODO: Works around https://github.com/dell/dkms/issues/177 for DKMS 3
-    echo "* adding hid-xpadneo-${VERSION} folder to /usr/src"
-    mkdir -p "/usr/src/hid-xpadneo-${VERSION}"
-    cp --recursive "${V[@]}" hid-xpadneo/. "/usr/src/hid-xpadneo-${VERSION}/."
+    cp -a hid-xpadneo/ /usr/src/hid-xpadneo-${VERSION}
+    dkms add -v "hid-xpadneo/${VERSION}" --force || cat_dkms_make_log
```
kakra, line 17: Looking at the build logs, the parameter `-v` seems to be conflicting. I'm not sure why you added it.

PR #365, opened 2022-05-30
Map the Share button to F12

kakra:


orbea:
This may conflict with other programs that use the `F12` button for other reasons?

kakra:
> This may conflict with other programs that use the `F12` button for other reasons?

That may well be but OTOH, the current mapping cannot even be assigned in many software. So v0.10.x will ship with this as the default while the development branch to v0.11 will add customization.

PR #381, opened 2022-08-11
Remove outdated secure boot docs

### LINKED ISSUE #380: Secure Boot problem
HEYYY your instruction for secure boot is nice
but where are those scripts?
in misc there no module-signing directory

HandyHat:
From what I can tell, these instructions are unnecessary since https://github.com/atar-axis/xpadneo/pull/71, which is the PR that removed the scripts these docs are referencing.

Closes #380


kakra (commented):
I never used module signing but from some community discussions it looks like module signing just works in distributions supporting it. I think it's okay then to remove these docs? IOW, can you confirm that observation?

Also, please put your initial comment into the commit message.

HandyHat:
> Also, please put your initial comment into the commit message.

Have done :)

PR #396, opened 2022-11-23
README.md: Fedora instructions are incorrect

jbillingredhat:
Fedora does not build a `kernel-headers` package for each release of the kernel, so you can't just use `uname -r` to install it.  Installing just the `kernel-headers` package is sufficient.  This is slightly confusing to people coming from other distros.  The `kernel-headers` package only includes files used to build userspace software, so it doesn't need to track the kernel closely.

The `kernel-devel` does track the kernel, so that is appropriate to install using `uname -r`.


kakra (approved):
LGTM

PR #412, opened 2023-03-17
Use --depth=1 to reduce download size

Apacelus:
Reduces the download size from ~23.4mb to ~3mb

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -235,7 +235,7 @@ Please feel free to add other distributions as well!
 ### Installation
 
 * Download the Repository to your local machine
-  `git clone https://github.com/atar-axis/xpadneo.git`
+  `git clone --depth=1 https://github.com/atar-axis/xpadneo.git`
 * `cd xpadneo`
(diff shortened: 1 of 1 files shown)


kakra:
I feel like this would cause more harm than benefit. People are encouraged to install via a distribution package. Changing the clone depth would prevent people from properly using branches when asked for in the issues. So I'd argue that people who got that far down are more likely to do development or issue reporting stuff. A size delta of 20 MB is just too small to outperform the possible drawbacks. Yes, it's like 85% less storage but in absolute numbers it's just 20 MB - that's a nothingness in today's systems.

If anything like this should go into the documentation, it should be for the stable branch. Unfortunately, this project doesn't have a linear stable branch, we branch off the current development branch when a development feature is ready and backport patches much like the kernel does. This would mean to change the docs on each stable cycle to the new stable branch name.

I'll leave this open as a reminder to amend the docs to encourage people to use their distribution package or a stable release.

PR #417, opened 2023-04-14
Prepare release v0.9.6

kakra:


SpidFightFR:
heya any update on this ? :sweat_smile:

kakra:
Oh well, I should really follow up on this... But time is very short currently.

SpidFightFR:
> Oh well, I should really follow up on this... But time is very short currently.

haha no pressure !
It's just that for the Voidlinux distribution (doesn't support git packages like arch do), xpadneo commits post v0.9.5 brought some fixes that made the xbox series controller work on my side (i had arch before void, so i noticed `xpadneo` from the AUR had the same issue as my current void install, unlike `xpadneo-git`). 

I subbed to the PR anyways, so when v0.9.6 will be up, then i'll update the void package.

PR #421, opened 2023-04-25
hid-xpadneo, Makefile: remove requirement for kernel sources

sewnie:
remove the need for needing kernel sources for building and installing xpadneo. when i first found this project i was very weirded by the fact that it required the kernel sources, when many modules i know only need the header files, or modules directory.

i can add the ability to use kernel sources if required, automatically via sourching kernel's Kbuild.

i've also changed it so that `VERSION` isn't a requirement, and can be set at runtime `make VERSION=bleh`, for package builds that don't have access to a git directory.

### DIFF
--- a/hid-xpadneo/Makefile
+++ b/hid-xpadneo/Makefile
@@ -1,5 +1,7 @@
-KERNEL_SOURCE_DIR := /usr/src/linux
-LD := ld.bfd
+ifndef KERNEL_SOURCES
+	KERNEL_UNAME ?= $(shell uname -r)
+	KERNEL_SOURCES := /lib/modules/$(KERNEL_UNAME)/build
+endif
 
 all: modules
 
@@ -9,7 +11,7 @@ all: modules
 	git describe --tags --dirty >$@
 
 clean modules modules_install: ../VERSION
-	$(MAKE) -C $(KERNEL_SOURCE_DIR) INSTALL_MOD_DIR="kernel/drivers/hid" LD=$(LD) M=$(shell pwd)/src VERSION="$(shell cat ../VERSION)" $@
+	$(MAKE) -C $(KERNEL_SOURCES) INSTALL_MOD_DIR="kernel/drivers/hid" M="$(shell pwd)/src" VERSION=$(shell cat ../VERSION) $@
 
 reinstall: modules
 	sudo make modules_install


hid-xpadneo/Makefile
```diff
@@ -1,15 +1,13 @@
-KERNEL_SOURCE_DIR := /usr/src/linux
-LD := ld.bfd
+KERNEL_UNAME  ?= $(shell uname -r)
+KERNEL_BUILD := /lib/modules/$(KERNEL_UNAME)/build
```
kakra, line 2: There's a patch in my queue already for this but it didn't work reliably across different environments. But I can push it so you could rebase your work and see if you still need changes.

Tekudachi:
Hi, how can I sign this driver for enabling secure boot? Thanks in advance.

hid-xpadneo/Makefile
```diff
@@ -1,15 +1,13 @@
...
-all: modules
-
-.INTERMEDIATE: ../VERSION
+VERSION ?= $(shell git describe --tags --dirty)
 
-../VERSION:
-	git describe --tags --dirty >$@
+all: modules
 
-clean modules modules_install: ../VERSION
```
kakra, line 11: This cannot be removed, it is actually needed for the rest of the scripts to work. The other script generate this file before calling this make file, it just depends on the existence of the file. It is not supported to build the source without setting the version but your change allows this. People not having git, will have downloaded a versioned archive which includes this file. If you're using custom archives without version file, please adjust your build scripts to first generate this file.

I would accept a change which makes the presence of the file optional if you pass a `VERSION` variable. In this case, the file should still be created as an intermediate target so it becomes deleted after executing make.

kakra (changes requested):

hid-xpadneo/Makefile
apprehensions: ?

hid-xpadneo/Makefile
apprehensions: so instead of running a command you depend on a file? why was that that preferred?

apprehensions: on-top of that, wouldn't those scripts require some other form of build script to have been executed first to generate that file?

kakra:
> Hi, how can I sign this driver for enabling secure boot? Thanks in advance.

This is specific to each distribution so it was removed from the docs some while ago (maybe there are a few leftover generic instructions). Usually, your distribution DKMS should take care of it. If it does not, you should ask your distribution maintainers. If this needs an optional hook in the dkms config, I can add that. Please create a new issue for this or use one of the reports that already handle that topic. Thanks. :-)

hid-xpadneo/Makefile
kakra: The downloadable archives are distributed with the version file included. It is generated when I tag a new version. But when running from a git repo, the file is generated dynamically instead from `git describe`. So it works for both cases without too much trouble in the `Makefile` itself. Thus, using an intermediate target to have a version file (either generated on the fly, or statically put there by a package maintainer or by an extracted archive) always works the same within the build script.

So, no, there's no requirement for a build script to generate the file: It is either distributed with the official release tarball, or it is extracted from git (either for the master branch or by checking out a tag). Using a non-release tarball is not supported because it easily led to incorrect version numbers in the past.

If you're using a non-release tarball (e.g., a git archive export) with possible custom patches and without git, you're required to generate a proper version file reflecting your own patch level and customization. This is similar to how `autoconf` projects ship release tarballs with pre-generated files but if you download a non-release tarball or checkout from git, you're required to generate these files by yourself.

hid-xpadneo/Makefile
kakra: I'll push my changes later this day. Your patch confirms, that I should really push it and no longer let it sit in my local queue:

https://github.com/kakra/xpadneo/commit/216f094052fd0c7015b3c0421a45dc2f5cc958df

It currently uses the `source` symlink but that's more an artifact of my local build environment. I'll push it with a `build symlink`.

hid-xpadneo/Makefile
apprehensions: is 87665bb9ed99e47d95aea8b1493b7a30e04c1b91 sufficient? or does it still need a file

kakra: > is [87665bb](https://github.com/atar-axis/xpadneo/commit/87665bb9ed99e47d95aea8b1493b7a30e04c1b91) sufficient? or does it still need a file

This still doesn't respect that a version file shipped with the release tarballs exists. You'd probably need to change the rule which creates the intermediate file instead of completely eliminating it. But it would be much easier, if you just created that file before calling make.

May I ask why you are using a source tree without version file and without `.git` in the first place?

apprehensions: my apologies. 

> May I ask why you are using a source tree without version file and without .git in the first place?

in some packaging scenarios, a source package is heavily preferred, such as a tarball from a given commit, where a .git file would not exist; as its provided from a git clone. which already use DKMS, and set the `@DO_NOT_CHANGE@` variable manually, such as void-packages.

apprehensions: using a version file will not allow removing it to begin with, as it is going to be forced, so one can't choose between using a variable or a file.

apprehensions: since i unfortunately have no choice, the only thing the user; to make the version file optional and to set it at build time, is to make the file themselves.

kakra: For this scenario, use a stable source tarball: https://github.com/atar-axis/xpadneo/archive/refs/tags/v0.9.5.tar.gz

The "master" branch is not meant to be used without git metadata. If needed, you'd need to create a version file, e.g. `echo "v0.9-void-${commit_id}" >VERSION`, which should be possible because you write "given commit". Otherwise, use stable tags instead of a git commit id.

If it makes things easier for you, I can make a stable branch which always has the latest version no matter the branch version.

The source code is meant to be built by users from a git clone, and they are not expected to guess the version number and provide it on command line. As an alternative, stable tarballs can be used, which ship no `.git` but a `VERSION` file. You proposed change breaks both expectations. As a package maintainer, it should be easy to just create the version file as part of your build script because your package has version metadata if you really want to build from arbitrary commit ids.

Arch handles it in the packaging script: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=xpadneo-dkms (using `sed`)

Gentoo does, too: https://gitweb.gentoo.org/repo/gentoo.git/tree/games-util/xpadneo/xpadneo-0.9.5.ebuild (using package metadata)

PR #441, opened 2023-10-24
Update BT_DONGLES

liberodark:
Add Missing Dongles


puleglot:
UB500 is based on Realtek RTL8761B* chip

PR #456, opened 2024-01-29
Add BUILD_EXCLUSIVE_CONFIG to dkms.conf

kevinoid:
Since version 3.0.11 (<https://github.com/dell/dkms/pull/269>), dkms supports the [`BUILD_EXCLUSIVE_CONFIG`] option in dkms.conf to specify kernel configuration options which must be enabled in the targeted kernel.  Add this option with `CONFIG_HID`, `CONFIG_INPUT_FF_MEMLESS`, and `CONFIG_POWER_SUPPLY`, since xpadneo will fail to modpost if any of these are disabled.  If a user installs a kernel where these options are disabled, the module will not be built and a warning similar to the following is printed:

    Warning: The /var/lib/dkms/hid-xpadneo/v0.9-144-g9b3b696-dirty/6.8.0-rc2/x86_64/dkms.conf
    for module hid-xpadneo includes a BUILD_EXCLUSIVE directive
    which does not match this kernel/arch/config.
    This indicates that it should not be built.
    dkms autoinstall on 6.8.0-rc2/x86_64 was skipped for hid-xpadneo

Without this option, the module build would fail and the kernel package would fail to install, leaving the user with a partially installed kernel package (on Debian, at least).

Thanks for considering,
Kevin

[`BUILD_EXCLUSIVE_CONFIG`]: https://github.com/dell/dkms/blob/782bd07641b957012f9b8d621bf9b51f82bfc728/dkms.8.in#L688-L696


kakra (approved):
LGTM, will merge.

PR #471, opened 2024-04-21
xpadneo: Fix missing report check

yjun123:
This fix ensures that the missing_reported flag is correctly updated when the keyboard or gamepad is not detected.


kakra (approved):
Good find! Thanks a lot... :-)

PR #488, opened 2024-08-02
bluez-tools isn't required on Tumbleweed

mscharley:
Just a quick update to remove a package that doesn't exist anymore: https://software.opensuse.org/search?q=bluez-tools&baseproject=openSUSE%3AFactory

I was able to install the kernel module and play with my controller just fine with no substitute, I believe these tools have been rolled into the main `bluez` package now.


kakra (approved):
LGTM

PR #493, opened 2024-09-04
docs: UGREEN CM390 compatibility status and some other changes

puleglot:


docs/BT_DONGLES.md
```diff
@@ -102,20 +102,29 @@ If `btmgmt` command is not available, try `bluetoothctl mgmt.info` instead.
...
+
+Known bad firmware for RTL8761BU chipset is 0xdfc6d922. It causes frequent
+reconnects. Firmware version can be found in the kernel log:
+```
+$ sudo dmesg | grep 'RTL: fw version'
+[   21.193448] Bluetooth: hci0: RTL: fw version 0xdfc6d922
+```
+
 * [TP-Link USB Bluetooth Adapter Bluetooth 5.0 (UB500)](https://www.amazon.com/gp/product/B09DMP6T22)
-  * Chipset: CSR ???
```
kakra, line 106: Are you sure this is a Realtek chipset instead of CSR? I wonder if we may see multiple variants of the same product.

docs/BT_DONGLES.md
```diff
@@ -20,6 +20,7 @@ hci0:   Primary controller
         name jupiter
         short name
 ```
+If `btmgmt` command is not available, try `bluetoothctl mgmt.info` instead.
```
kakra, line 23: LGTM, but the commit subject should read "command" instead of "commant".

kakra (changes requested):
There are minor nitpicks and a few things to discuss/sort out...

docs/BT_DONGLES.md
puleglot: Well, if it was CSR, then it would definitely have different USB ID or it would not work with the linux kernel:
https://github.com/torvalds/linux/blob/0c3836482481200ead7b416ca80c68a29cfdaabd/drivers/bluetooth/btusb.c#L716

Also I haven't seen any CSR based BT 5.0 dongles. Maybe they don't have a suitable chipset at all as the company was bought by Qualcomm long ago.

kakra: Okay, makes sense. Looking back in my grey memory, there may have been some confusion about the UB400 vs UB500. Grouping by chipset manufacturer thus looks fine, although I wonder if - by end user perspective - it would be better to group by product manufacturer. But that should be a concern of another change. I'll take this.

Thanks. :-)

kakra (approved):
LGTM

Issue #484, opened 2024-07-27
Vibration not work on Gamesir T4 Nova Lite


BakaJzon:
## Version of xpadneo

xpadneo-dkms-git & xpadneo-dkms 0.9.6-1 from AUR
`-git` means using the latest commit on master branch to build

## Controller Model

- [ ] Xbox One S controller
- [ ] Xbox Elite 2 controller
- [ ] Xbox Series X|S controller
- [x] Other: Gamesir T4 Nova Lite or 盖世小鸡启明星

### Connection mode

- [x] Bluetooth connection
- [ ] USB cable (not yet supported)
- [ ] Xbox Dongle connection (not yet supported)

## Installed Software

- [ ] Anti-Micro (may affect button mappings)
- [ ] OpenRGB (may mess up mappings and rumble stability)
- [x] Steam Input (enabled by default via Steam Desktop client)
- [ ] Steam Link (usually via Raspberry Pi or other micro computers)
- [ ] devices with QMK firmware (may affect udev rules, similar to OpenRGB)
- [ ] netstick (shares input devices via network similar to Steam Link)
- [ ] xboxdrv (user-space gamepad driver)
- [ ] xone (kernel-space gamepad driver using the Xbox dongle or USB)
- [ ] xow (alternative driver using the Xbox dongle)

## Protocol Information

Please help us identify at which layer the problem can be found if you want
to report mapping errors or if the controller fails to be detected:

- [x] Steam Proton games are having issues
- [x] Steam Linux-native games are having issues
    - [ ] I don't use Steam or did not try
- [x] games running through Lutris, wine and/or Bottles are having issues
    - [ ] I don't use Lutris, Bottles, wine or did not try
- [x] Linux-native games are having issues
    - [ ] I don't use native games or did not try
- [ ] Other software is having issues (describe software and issues below)
- [ ] Running `evtest` is showing issues (describe the issues below)
    - Keep in mind that `BTN_NORTH` and `BTN_WEST` are intentionally swapped
- [ ] Running `jstest` is showing issues (describe the issues below)
    - [x] I don't have this tool or don't know how to use it
- [ ] Running `gamepad-tool` is showing issues (post console output below)
    - [x] I don't have this tool

Please describe how it is failing below in the next sections.

## Severity / Impact

- [x] I've read the docs and the bug reporting instructions
- [x] I've applied the latest firmware update to the controller
- [x] I've tried disabling or running without above mentioned software
- [ ] It does not work at all
[...]

PR #496, opened 2024-09-22
core, quirks: Add GameSir T4 Nova Lite support

kakra:
This family of controllers doesn't support trigger rumble but still needs all the motor-enable bits set unconditionally.

Fixes: https://github.com/atar-axis/xpadneo/issues/484

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -191,6 +191,21 @@ options hid_xpadneo quirks=A0:5A:5D:xx:xx:xx+2
 This controller uses emulated profile switching support (see below).
 
 
+### GameSir T4 Nova Lite Family
+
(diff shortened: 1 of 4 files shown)


SherLock707:
Would this work for dongle connection?

kakra:
> Would this work for dongle connection?

Only if the dongle presents the device in HID mode... xpadneo is a HID driver, it doesn't care about the connection, only about the protocol. That said, many dongles either present the device in USB input mode (no longer HID), or they pretend to be a very different kind of controller and no longer an Xbox controller. YMMV.

SherLock707:
In dongle connection right now and light is yellow.

`Bus 001 Device 014: ID 3537:1041 Zikway HID gamepad`

PR #550, opened 2025-07-03
xpadneo, docs: Update links to other projects

l3d00m:
* Use dlundqvist xone fork (which is also used on AUR), since the original one by medusalix is not active anymore and also doesn't build on newer kernels
* Mark xow as unmaintained since it hasn't received updates for many years and also doesn't build on newer kernels


kakra (approved):
LGTM

PR #559, opened 2025-09-19
(fix): change rules to match systemd v258 onwards

Edu4rdSHL:
This change is compatible with previous versions as well. See https://github.com/systemd/systemd/blob/781d9d0789379d1ea1f2ecefb804d41e9c8b6c38/NEWS#L22-L32

### DIFF
--- a/hid-xpadneo/etc-udev-rules.d/60-xpadneo.rules
+++ b/hid-xpadneo/etc-udev-rules.d/60-xpadneo.rules
@@ -2,4 +2,4 @@
 ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", KERNEL=="0005:045E:*", KERNEL=="*:02FD.*|*:02E0.*|*:0B05.*|*:0B13.*|*:0B20.*|*:0B22.*", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
 
 # Tag xpadneo devices for access in the user session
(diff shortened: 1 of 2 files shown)


Edu4rdSHL:
> How is this backwards compatible with older versions?

Hi, thanks for your prompt answer. I work closely with the systemd developers, and the current approach where people were using `add` is incorrect. With the current change, the issue arose on several applications, and we are trying to prevent further breakage. The `!=` condition has always worked; it's `=="add"` that stopped working, but since we have `"add|change"`, the correct approach is `!="remove"`.

kakra:
> > How is this backwards compatible with older versions?
> 
> Hi, thanks for your prompt answer. I work closely with the systemd developers, and the current approach where people were using `add` is incorrect. With the current change, the issue arose on several applications, and we are trying to prevent further breakage. The `!=` condition has always worked; it's `=="add"` that stopped working, but since we have `"add|change"`, the correct approach is `!="remove"`.

Okay, thanks for clarification.

Is it safe to use with LTS distributions which ship (possibly much) older versions of systemd? If yes, I would consider it for merging now.

Edu4rdSHL:
Yes, it's safe to use with LTS distros. At the functional level, `ACTION!="remove"` is only slightly different from `ACTION=="add|change"` in the fact that it will catch a couple more events, but that's not a problem for tagging devices with `uaccess`, which is where the change happens.

LTS distros will continue working as of now, but with better event handling and proper compatibility with systemd 258+ as well.

PR #563, opened 2025-10-11
Add scripts that sign the module for Secure Boot

flodavid:
I created scripts to automate the signing and installation steps required on systems with Secure Boot enabled

- For DKMS: /sign_and_install.sh checks that files configured as MOK for DKMS exist and the MOK is enrolled. If it is not the case, then it creates and/or enroll them using either Ubuntu’s tool or `openssl` (or `dkms`) and `mokutil` commands.
Then it run the install script that should build and sign the module using standard DKMS procedure.
- For generic: /hid-xpadneo/sign_and_install_generic.sh applies the same checks, but on “/var/lib/shim-signed/mok” as MOK files location.
Then it build and install the module using make commands.
- /lib/mok.sh handles MOK checks, creation and enrollment.
- /hid-xpadneo/Makefile has been modified to support signing after building the module. It uses MOK located in “/var/lib/shim-signed/mok”.
- /Makefile is modified to better clean files on uninstall.

- [x]  My contribution has been created with the help of AI, including prior research with AI chatbots or AI-assisted
reviews and suggestions (e.g. GitHub Copilot PR suggestions). I will not contribute commits with a
Co-authored-by signature of an AI agent because I confirm that my contribution has been verified, reviewed and tested by a human being, and I take full responsibility as the author of the commit using my own Signed-off-by.

(The only AI contribution is the Github Copilot review below)

### DIFF
--- a/hid-xpadneo/sign_and_install_generic.sh
+++ b/hid-xpadneo/sign_and_install_generic.sh
@@ -1,0 +1,66 @@
+#!/usr/bin/env bash
+#set -x
+set -e
+
+# https://docs.fedoraproject.org/en-US/fedora/rawhide/system-administrators-guide/kernel-module-driver-configuration/Working_with_Kernel_Modules/#sect-signing-kernel-modules-for-secure-boot
+
+source "$(dirname "$0")/../lib/mok.sh"
+
+if [ "$EUID" -ne 0 ]
+then error "Please run as root"
+fi
+
+command -v openssl > /dev/null 2>&1 \
+  || warn "missing openssl, you will not be able to create a new MOK key"
+command -v mokutil > /dev/null 2>&1 \
+  || warn "missing mokutil, you will not be able to sign the module"
+
+function install_xpadneo () {
+  info "start install_xpadneo"
+
+  make reinstall-sign
+  # Generic install already loads the module, so no need to use depmod and modprobe
+
+  if modinfo hid-xpadneo
+  then
+    info "finished install_xpadneo"
+
+    if [ -d /sys/module/hid_xpadneo ]
+    then
+      info "hid_xpadneo loaded!"
+      # try to load uhid to support Bluetooth LE (for firmware 5.x)
+      if [ ! -d /sys/devices/virtual/misc/uhid ]
+      then modprobe uhid
+      fi
+    else warn "failed to load hid_xpadneo (check for errors by running 'sudo dmesg')"
+    fi
+  else warn "failed to load hid_xpadneo"
+  fi
+
+  info "finished install_xpadneo"
+}
+
+if [ "$#" -ne "0" ]
+then
+  signing_help
+  exit 0
+else
+  if mokutil --sb-state | grep -q "SecureBoot enabled"; then
+    if ! [ -d /sys/module/hid_xpadneo ]
+    then
+      info "Secure boot is enabled, signing is needed"
+      # Check/enroll keys in default shim directory
+      if check_keys "/var/lib/shim-signed/mok"
+      then install_xpadneo
+      fi
+    else
+      info "hid_xpadneo is loaded"
+    fi
+  else
+    warn "Secure boot is disabled, no need to sign module!"
+    export skip_signing=1
+    if ! [ -d /sys/module/hid_xpadneo ]
+    then install_xpadneo
+    fi
+  fi
+fi
--- a/sign_and_install.sh
+++ b/sign_and_install.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+#set -x
+set -e
+
+# https://docs.fedoraproject.org/en-US/fedora/rawhide/system-administrators-guide/kernel-module-driver-configuration/Working_with_Kernel_Modules/#sect-signing-kernel-modules-for-secure-boot
+
+source "$(dirname "$0")/lib/mok.sh"
+
+command -v openssl > /dev/null 2>&1 \
+  || warn "missing openssl, you will not be able to create a new MOK key"
+command -v mokutil > /dev/null 2>&1 \
+  || warn "missing mokutil, you will not be able to sign the module"
+
+function install_xpadneo () {
+  info "start install_xpadneo"
+
+  # Uninstall previous version
+  source "$(dirname "$0")/uninstall.sh"
+  source "$(dirname "$0")/install.sh"
+
+  module_ko_file=$(modinfo hid-xpadneo | awk '/^filename:/ { print $2 }')
+  if [ -z "$module_ko_file" ]
+  then error "Module compilation or installation failed"
+  fi
+
+  depmod -a
+
+  if modprobe hid-xpadneo && modinfo hid-xpadneo
+  then
+    info "finished install_xpadneo"
+
+    if [ -d /sys/module/hid_xpadneo ]
+    then
+      info "hid_xpadneo loaded!"
+      # try to load uhid to support Bluetooth LE (for firmware 5.x)
+      if [ ! -d /sys/devices/virtual/misc/uhid ]
+      then modprobe uhid
+      fi
+    else warn "failed to load hid_xpadneo (check for errors by running 'sudo dmesg')"
+    fi
+  else warn "failed to load hid_xpadneo"
+  fi
+}
+
+if [ "$#" -ne "0" ]
+then
+  signing_help
+  exit 0
+else
+  if mokutil --sb-state | grep -q "SecureBoot enabled"
+  then info "Secure boot is enabled, signing is needed"
+  else warn "Secure boot is disabled, no need to sign module!"
+  fi
+
+  if ! [ -d /sys/module/hid_xpadneo ]
+  then
+    # Check in DKMS directory and enroll if a MOK is present
+    info "If you want to use specific key, set mok_signing_key and mok_certificate in " \
+     "/etc/dkms/framework.conf. Otherwise, shim default will be used."
+    if check_keys ""
+    then install_xpadneo
+    fi
+  else info "hid_xpadneo is already loaded"
+  fi
+fi
--- a/docs/README.md
+++ b/docs/README.md
@@ -376,6 +376,17 @@ Please feel free to add other distributions as well!
 - If not using DKMS, follow steps above (generic distribution)
 - Done!
 
+#### Sign the module
+
+If Secure Boot is enabled on your machine, your system will not accept to load the unsigned module
+and you may get this kind of error:
+`modprobe: ERROR: could not insert 'hid_xpadneo': Key was rejected by service`.
+You need to sign it:
+
+* If using DKMS, run `sudo ./sign_and_install.sh` in lieu of `install.sh`.
+* For generic distribution, run `cd hid-xpadneo && sudo ./sign_and_install_generic.sh`
+
+You may need to run the scripts twice, with a reboot in-between, according to the output.
 
 ### Connection
 
--- a/hid-xpadneo/Makefile
+++ b/hid-xpadneo/Makefile
@@ -1,9 +1,25 @@
 KVER ?= $(shell uname -r)
+
 INSTALL_MOD_DIR := kernel/drivers/hid
 KERNEL_SOURCE_DIR ?= /lib/modules/$(KVER)/build
 KERNEL_DEST_DIR := /lib/modules/$(KVER)/$(INSTALL_MOD_DIR)
 MOK_DIR ?= /var/lib/shim-signed/mok
+ETC_PREFIX ?= /etc
+DOC_PREFIX ?= /usr/share/doc/xpadneo
+META_PREFIX ?= /usr/share/metainfo
+
 LD := ld.bfd
+MODPROBE_CONFS := xpadneo.conf
+UDEV_RULES := 60-xpadneo.rules 70-xpadneo-disable-hidraw.rules
+DOC_SRCS := ../NEWS.md $(wildcard ../docs/[0-9A-Z]*.md)
+DOCS := $(notdir $(DOC_SRCS))
+
+ifeq ($(PREFIX),)
+  UDEVADM ?= udevadm
+else
+  $(warning Installing to prefix, udevadm commands will not be run!)
+  UDEVADM ?= : SKIPPING udevadm
+endif
 
 all: modules
 
@@ -20,8 +36,28 @@ clean modules modules_install: ../VERSION
 check_sudo:
 	[ "$(shell id -u)" = "0"  ] || { echo >&2 "needs to be run with sudo"; exit 1; }
 
-reinstall: modules check_sudo modules_install unload load
-reinstall-sign: modules check_sudo modules_install sign unload load
+reinstall: modules check_sudo install-all unload load
+reinstall-sign: modules check_sudo install-all sign unload load
+install-all: install install-metainfo
+
+install: modules_install
+	mkdir -p $(PREFIX)$(ETC_PREFIX)/modprobe.d $(PREFIX)$(ETC_PREFIX)/udev/rules.d $(PREFIX)$(DOC_PREFIX)
+	install -D -m 0644 -t $(PREFIX)$(ETC_PREFIX)/modprobe.d $(MODPROBE_CONFS:%=etc-modprobe.d/%)
+	install -D -m 0644 -t $(PREFIX)$(ETC_PREFIX)/udev/rules.d $(UDEV_RULES:%=etc-udev-rules.d/%)
+	install -D -m 0644 -t $(PREFIX)$(DOC_PREFIX) $(DOC_SRCS)
+	$(UDEVADM) control --reload
+
+install-metainfo:
+	install -D -m 0644 -t $(PREFIX)$(META_PREFIX) ../io.github.atar_axis.xpadneo.metainfo.xml
+
+uninstall: check_sudo unload ../VERSION
+	rm -f $(KERNEL_DEST_DIR)/hid-xpadneo.ko*
+	rm -f $(DOCS:%=$(PREFIX)$(DOC_PREFIX)/%)
+	rm -f $(UDEV_RULES:%=$(PREFIX)$(ETC_PREFIX)/udev/rules.d/%)
+	rm -f $(MODPROBE_CONFS:%=$(PREFIX)$(ETC_PREFIX)/modprobe.d/%)
+	rm -f $(PREFIX)$(META_PREFIX)/io.github.atar_axis.xpadneo.metainfo.xml
+	rmdir --ignore-fail-on-non-empty -p $(PREFIX)$(ETC_PREFIX)/modprobe.d $(PREFIX)$(ETC_PREFIX)/udev/rules.d $(PREFIX)$(DOC_PREFIX) || true
+	$(UDEVADM) control --reload
 
 $(MOK_DIR)/MOK.priv $(MOK_DIR)/MOK.der:
 # Create the shim MOK keys if they do not exist, using update-secureboot-policy.
@@ -48,6 +84,7 @@ unload: check_sudo
 	rmmod hid-xpadneo || echo "proceeding anyways"
 
 load: check_sudo
+	depmod -a
 	modprobe hid-xpadneo $(MOD_PARAMS)
 
 # DKMS support rules
--- a/hid-xpadneo/Makefile
+++ b/hid-xpadneo/Makefile
@@ -1,4 +1,8 @@
-KERNEL_SOURCE_DIR ?= /lib/modules/$(shell uname -r)/build
+KVER ?= $(shell uname -r)
+INSTALL_MOD_DIR := kernel/drivers/hid
+KERNEL_SOURCE_DIR ?= /lib/modules/$(KVER)/build
+KERNEL_DEST_DIR := /lib/modules/$(KVER)/$(INSTALL_MOD_DIR)
+MOK_DIR ?= /var/lib/shim-signed/mok
 LD := ld.bfd
 
 all: modules
@@ -11,12 +15,40 @@ all: modules
 # convenience rules for local development
 
 clean modules modules_install: ../VERSION
-	$(MAKE) -C $(KERNEL_SOURCE_DIR) INSTALL_MOD_DIR="kernel/drivers/hid" LD=$(LD) M=$(shell pwd)/src VERSION="$(shell cat ../VERSION)" $@
+	$(MAKE) -C $(KERNEL_SOURCE_DIR) INSTALL_MOD_DIR="$(INSTALL_MOD_DIR)" LD=$(LD) M=$(shell pwd)/src VERSION="$(shell cat ../VERSION)" $@
 
-reinstall: modules
-	sudo make modules_install
-	sudo rmmod hid-xpadneo || true
-	sudo modprobe hid-xpadneo $(MOD_PARAMS)
+check_sudo:
+	[ "$(shell id -u)" = "0"  ] || { echo >&2 "needs to be run with sudo"; exit 1; }
+
+reinstall: modules check_sudo modules_install unload load
+reinstall-sign: modules check_sudo modules_install sign unload load
+
+$(MOK_DIR)/MOK.priv $(MOK_DIR)/MOK.der:
+# Create the shim MOK keys if they do not exist, using update-secureboot-policy.
+# Users have to handle it themselves if it is not installed or not a good version.
+ifeq ($(shell update-secureboot-policy --help 2>&1 | grep 'new-key'),)
+	@echo "update-secureboot-policy not installed or does not have 'new-key' option. You will have to create the MOK yourself in $(MOK_DIR)"
+else
+	make check_sudo
+	@update-secureboot-policy --new-key
+endif
+
+sign: $(MOK_DIR)/MOK.priv $(MOK_DIR)/MOK.der $(KERNEL_DEST_DIR)/hid-xpadneo.ko
+	make check_sudo
+# Enroll the shim MOK public key
+ifeq ($(shell update-secureboot-policy --help 2>&1 | grep 'enroll-key'),)
+	mokutil --import "$(MOK_DIR)/MOK.der"
+else
+	update-secureboot-policy --enroll-key | grep -q --invert-match 'No DKMS modules installed' \
+	  || mokutil --import "$(MOK_DIR)/MOK.der"
+endif
+	$(KERNEL_SOURCE_DIR)/scripts/sign-file sha256 $^
+
+unload: check_sudo
+	rmmod hid-xpadneo || echo "proceeding anyways"
+
+load: check_sudo
+	modprobe hid-xpadneo $(MOD_PARAMS)
 
 # DKMS support rules
 
--- a/lib/mok.sh
+++ b/lib/mok.sh
@@ -1,0 +1,189 @@
+#!/bin/bash
+# set -x
+
+DKMS_DEFAULT_MOK_DIR="/var/lib/dkms"
+UBUNTU_DEFAULT_MOK_DIR="/var/lib/shim-signed/mok"
+USP_MOK_DIR="/var/lib/shim-signed/mok"
+
+if [ ${EUID} -ne 0 ]
+then
+	echo >&2 "ERROR: You most probably need superuser privileges to use this script, \
+please run me via sudo!"
+	exit 3
+fi
+
+function info  { echo -e "\e[32m[info] $*\e[39m"; }
+function warn  { echo -e "\e[33m[warn] $*\e[39m"; }
+function error { echo -e "\e[31m[error] $*\e[39m"; exit 1; }
+
+function signing_help () {
+  echo "
+    just run the script, reboot (you will have to type a pass you set) and run again!
+  "
+}
+
+# Check that Machine Owner Keys exist and are enrolled. Generate new ones if necessary.
+# Accept a parameter for MOK location to check. Otherwise, use configured or default DKMS directory.
+# If it exists, enroll MOK if necessary, but do not create new ones.
+# If no key file found, we generate MOK.
+# Return 0 if key already exists and is enrolled, 1 if it needs a restart to enroll.
+function check_keys () {
+  moks_dir="$1"
+
+  # Load and use values set in DKMS (or Ubuntu) configuration file if no parameter
+  if [ -z "$moks_dir" ]
+  then
+    source /etc/dkms/framework.conf
+    [[ -r /etc/os-release ]] && source /etc/os-release
+
+    # DKMS uses the default key created by update-secureboot-policy for Ubuntu and the likes
+    if [[ -z "${mok_signing_key}" ]] \
+      && { [ "$ID" = "ubuntu" ] || [[ "${ID_LIKE[0]}" == "ubuntu"* ]]; }
+    then
+      echo "Using default MOK key location for Ubuntu and derivatives"
+      moks_dir="$UBUNTU_DEFAULT_MOK_DIR"
+    else
+      # Use MOK location defined in /etc/dkms/framework.conf. Fallback to default one
+      if [ -z "$mok_signing_key" ]
+      then mok_signing_key="$DKMS_DEFAULT_MOK_DIR/mok.key"
+      fi
+      if [ -z "$mok_certificate" ]
+      then mok_certificate="$DKMS_DEFAULT_MOK_DIR/mok.pub"
+      fi
+
+      echo "Using DKMS configured or default MOK files locations"
+      # Check if DKMS key already exists, create it otherwise
+      if [ -f "$mok_signing_key" ] && [ -f "$mok_certificate" ]
+      then
+        # If key is already trusted, no need to enroll it
+        if check_key "$mok_signing_key" "$mok_certificate"
+        then return 0
+        fi
+      else generate_dkms_mok "$mok_signing_key" "$mok_certificate"
+      fi
+
+      # MOK was just created or is not trusted yet
+      enroll_mok "$mok_certificate"
+      return 1
+    fi
+  fi
+
+  # Check if any key is present and enrolled
+  if check_key "$moks_dir/MOK.priv" "$moks_dir/MOK.der" \
+    || check_key "$moks_dir/mok.key" "$moks_dir/mok.pub"
+  then return 0
+  fi
+
+  info "No Private key found or missing associated DER file in $moks_dir. \
+Generating and importing it."
+  # Try to use update-secureboot-policy if the given directory corresponds to where it stores MOK
+  if [ "$USP_DEFAULT_MOK_DIR" = "$moks_dir" ]
+  then
+    generate_usp_mok
+    enroll_usp_mok
+  else
+    echo "Using custom MOK key location"
+    mkdir -p "$moks_dir"
+    generate_mok "$moks_dir/MOK.priv" "$moks_dir/MOK.der"
+    enroll_mok "$1/MOK.der"
+  fi
+  
+  return 1
+}
+
+# Used by check_keys to check MOK is valid
+# returns 0 if key pair exists and is valid, 1 if it does not exist or is not valid
+function check_key () {
+  if [ -f "$1" ] && [ -f "$2" ]
+  then
+    if mokutil --test-key "$2" | grep -q "already enrolled"
+    then
+      info "MOK Key $2 found, already trusted."
+      return 0
+    else
+      info "$2 MOK key exists but is not trusted yet. Importing it."
+      # Enrolling the MOK, even if not necessary
+      enroll_mok "$moks_dir"
+    fi
+  fi
+
+  return 1
+}
+
+# Used by check_keys to generate DKMS MOK files
+function generate_dkms_mok () {
+  # Use DKMS's generate_mok action, fallback on openssl
+  dkms_help_output=$(bash -c 'dkms --help' 2>&1)
+  if command -v dkms >/dev/null 2>&1 \
+    && echo "$dkms_help_output" | grep -q "generate_mok"
+  then
+    echo "Using 'dkms generate_mok' to manage MOK creation"
+    dkms generate_mok
+  else generate_mok "$1" "$2"
+  fi
+}
+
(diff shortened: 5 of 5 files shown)


hid-xpadneo/sign_and_install_generic.sh
```diff
@@ -0,0 +1,196 @@
...
+function import_trust () {
+  info "Enroling new cert"
+  warn "Set a one-time import password, Make it memorablel, you have to type it into later"
+  warn "for a example check https://gist.github.com/reillysiemens/ac6bea1e6c7684d62f544bd79b2182a4"
+  mokutil --import "${mokutil_out_dir}"/MOK.der
+  info "A Key has been generated and importet into mokutil"
+  mokutil --list-new
+  info "Now reboot your machine and import the certificate with the just typed password"
+  exit 0
+  info "finished import_trust"
```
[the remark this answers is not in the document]
flodavid: done

sign_and_install.sh
```diff
@@ -0,0 +1,195 @@
...
+function import_trust () {
+  info "Enroling new cert"
+  warn "Set a one-time import password, Make it memorablel, you have to type it into later"
+  warn "for a example check https://gist.github.com/reillysiemens/ac6bea1e6c7684d62f544bd79b2182a4"
+  mokutil --import "${mokutil_out_dir}"/MOK.der
+  info "A Key has been generated and importet into mokutil"
+  mokutil --list-new
+  info "Now reboot your machine and import the certificate with the just typed password"
+  exit 0
+  info "finished import_trust"
```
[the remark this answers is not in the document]
flodavid: done

hid-xpadneo/sign_and_install_generic.sh
```diff
@@ -0,0 +1,196 @@
...
+  chmod 600 "${mokutil_out_dir}"/.openssl_pass
+  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
+}
+
+function import_trust () {
+  info "Enroling new cert"
+  warn "Set a one-time import password, Make it memorablel, you have to type it into later"
+  warn "for a example check https://gist.github.com/reillysiemens/ac6bea1e6c7684d62f544bd79b2182a4"
+  mokutil --import "${mokutil_out_dir}"/MOK.der
+  info "A Key has been generated and importet into mokutil"
```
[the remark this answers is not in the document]
flodavid: done

sign_and_install.sh
```diff
@@ -0,0 +1,195 @@
...
+  chmod 600 "${mokutil_out_dir}"/.openssl_pass
+  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
+}
+
+function import_trust () {
+  info "Enroling new cert"
+  warn "Set a one-time import password, Make it memorablel, you have to type it into later"
+  warn "for a example check https://gist.github.com/reillysiemens/ac6bea1e6c7684d62f544bd79b2182a4"
+  mokutil --import "${mokutil_out_dir}"/MOK.der
+  info "A Key has been generated and importet into mokutil"
```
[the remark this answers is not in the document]
flodavid: done

hid-xpadneo/sign_and_install_generic.sh
```diff
@@ -0,0 +1,196 @@
...
+  cat "${mokutil_out_dir}"/MOK.pem > /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  #  cp "${mokutil_out_dir}"/MOK.der /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  chmod 600 "${mokutil_out_dir}"/MOK*
+  chmod 600 "${mokutil_out_dir}"/.openssl_pass
+  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
+}
+
+function import_trust () {
+  info "Enroling new cert"
+  warn "Set a one-time import password, Make it memorablel, you have to type it into later"
```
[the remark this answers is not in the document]
flodavid: done

hid-xpadneo/sign_and_install_generic.sh
```diff
@@ -0,0 +1,196 @@
...
+
+  cat "${mokutil_out_dir}"/MOK.pem > /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  #  cp "${mokutil_out_dir}"/MOK.der /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  chmod 600 "${mokutil_out_dir}"/MOK*
+  chmod 600 "${mokutil_out_dir}"/.openssl_pass
+  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
+}
+
+function import_trust () {
+  info "Enroling new cert"
```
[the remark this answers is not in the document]
flodavid: done

sign_and_install.sh
```diff
@@ -0,0 +1,195 @@
...
+  cat "${mokutil_out_dir}"/MOK.pem > /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  #  cp "${mokutil_out_dir}"/MOK.der /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  chmod 600 "${mokutil_out_dir}"/MOK*
+  chmod 600 "${mokutil_out_dir}"/.openssl_pass
+  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
+}
+
+function import_trust () {
+  info "Enroling new cert"
+  warn "Set a one-time import password, Make it memorablel, you have to type it into later"
```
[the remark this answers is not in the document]
flodavid: done

sign_and_install.sh
```diff
@@ -0,0 +1,195 @@
...
+
+function generate_certificate () {
+  openssl x509 -inform der -in "${mokutil_out_dir}"/MOK.der -out "${mokutil_out_dir}"/MOK.pem
+  mkdir -p /usr/src/kernels/"$(uname -r)"/certs/
+
+  cat "${mokutil_out_dir}"/MOK.pem > /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  #  cp "${mokutil_out_dir}"/MOK.der /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  chmod 600 "${mokutil_out_dir}"/MOK*
+  chmod 600 "${mokutil_out_dir}"/.openssl_pass
+  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
```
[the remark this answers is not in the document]
flodavid: done

hid-xpadneo/sign_and_install_generic.sh
```diff
@@ -0,0 +1,196 @@
...
+
+function generate_certificate () {
+  openssl x509 -inform der -in "${mokutil_out_dir}"/MOK.der -out "${mokutil_out_dir}"/MOK.pem
+  mkdir -p /usr/src/kernels/"$(uname -r)"/certs/
+
+  cat "${mokutil_out_dir}"/MOK.pem > /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  #  cp "${mokutil_out_dir}"/MOK.der /usr/src/kernels/"$(uname -r)"/certs/signing_key.pem
+  chmod 600 "${mokutil_out_dir}"/MOK*
+  chmod 600 "${mokutil_out_dir}"/.openssl_pass
+  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
```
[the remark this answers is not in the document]
flodavid: done

sign_and_install.sh
```diff
@@ -0,0 +1,195 @@
...
+}
+
+function import_trust () {
+  info "Enroling new cert"
```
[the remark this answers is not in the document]
flodavid: done

hid-xpadneo/sign_and_install_generic.sh
```diff
@@ -66,12 +65,15 @@ function signing_ko () {
     warn "there might be no file to sign. Please check ${ko_folder}"
     ko_folder="${ko_folder:-/lib/modules/$(uname -r)/updates}"
   fi
-  info "Cert Password: $mokutil_509key_pass"
+
   zstd -f --decompress "${ko_folder}/hid-xpadneo.ko.zst"
```
kakra: Can we be sure that modules will be always compressed? Or that it will always be zstd?

hid-xpadneo/sign_and_install_generic.sh
```diff
@@ -115,19 +117,20 @@ function generate_certificate () {
...
   chmod 600 "${mokutil_out_dir}"/.openssl_pass
-  echo "The pasword for the x509 cert is: ${mokutil_509key_pass}"
+  echo "The password for the x509 cert is: ${mokutil_509key_pass}"
 }
 
 function import_trust () {
-  info "Enroling new cert"
-  warn "Set a one-time import password, Make it memorablel, you have to type it into later"
+  info "Enrolling new cert"
+  warn "Set a one-time import password, Make it memorable, you have to type it into later"
```
kakra: If this is a one-shot action (generate, import), can we make the password generation automatic? Otherwise, I'd rather have the owner provide a previously generated cert they use for other modules, too... Maybe we could provide such a utility or provide documentation for it.

I didn't check how this blends into dkms rebuild during boot, tho. Will it ask for a password during auto rebuild?

sign_and_install.sh
```diff
@@ -162,7 +165,12 @@ function install_xpadneo () {
   modprobe hid-xpadneo
   modinfo hid-xpadneo
   info "finished install_xpadneo"
-  lsmod | grep -q hid_xpadneo && info "hid_xpadneo loaded!"
+
+  if lsmod | grep -q hid_xpadneo
```
kakra: I don't like the semantic of this check. Can't we check if modprobe itself failed or succeeded? Otherwise we don't see the result of `modprobe` here. We see the result of `lsmod` which doesn't reflect if the module was actually loaded previously, it only checks if the module is loaded - no matter the previous state.

If this is semantically okay, we should explicitly use `modprobe hid-xpadneo || true # ensure loaded` (or similar). If possible, the script should run properly with `set -e`.

sign_and_install.sh
```diff
@@ -186,7 +194,11 @@ then
   exit 0
 else
   if mokutil --sb-state | grep -q "SecureBoot enabled"; then
-    grep -q "hid_xpadneo" /proc/modules && info "hid_xpadneo is loaded" || ( info "Secure boot is enabled and you have to setup signing" ; check_key)
+    if ! grep -q "hid_xpadneo" /proc/modules && info "hid_xpadneo is loaded"
```
kakra: Why `lsmod` above but `/proc/modules` here? The difference is clear: It's more complicated to check with a pipe here. But then maybe we should create a function?

kakra (changes requested):
Mostly nitpicks. I appreciate your work but I have to admit that I did just a quick scan through your changes and didn't test it yet. That's because I don't run a secure boot system here. I'd need to setup a VM or container - probably with different distributions.

If that works, tho, I'm pretty confident I can use that as a template to make my bare metal system secureboot. So, thanks already here. :-)

hid-xpadneo/sign_and_install_generic.sh
flodavid: For another driver (new-lg4ff), the module was compressed with xz. I saw that the output was a zstd file there, so I used that (though for some reason, in /usr/… I had this same module uncompressed). It should work with an error message if a the module was not compressed. I can explicitly support an uncompressed module and add support for xz if you want

hid-xpadneo/sign_and_install_generic.sh
flodavid: Yeah it’s one shot, I used the same certificate for two modules. I did not found a standard location where to put this or a tool that handles it, but again I did not search a lot.
I am not familiar with dkms, but ~the motherboard~ `shim` MokManager is in charge of enrolling the key. So my guess is that loading the module would fail if there is an auto-rebuild and it was not manually signed again.
To avoid having to compile the whole module every time dkms auto-rebuild it, I could detect that it was modified and just re-sign it if possible (I could even ask the user if they want this to be checked on the session startup, with a notification when they need to re-sign it). I will check what’s possible when I have some time.

sign_and_install.sh
flodavid: Again no expert, but while I was trying to make this work,I found that `modprobe` can return errors (and you are right, I should check for those), but having no error did not seem to indicate that everything is up and running. I did not know if I should reconnect the controller , restart the Bluetooth or even restart to be sure. Since the script I adapted used the `lsmod` output, I guessed that they found this to be better way of making sure the installation succeeded, so I kept it. But I could be wrong.

sign_and_install.sh
flodavid: No idea to be honest, I will try to do some tests to see if the result is different or if one the check is better

sign_and_install.sh
kakra: Nothing that xpadneo does, affects the Bluetooth connection. If you load or unload the driver, the device stays connected because xpadneo doesn't talk to Bluetooth, it talks to the HID bus provided by the Bluetooth layer.

hid-xpadneo/sign_and_install_generic.sh
kakra: Usually, I expect that dkms itself handles this. So as long as this does not interfere with that automatic behavior, we should be fine. OTOH, I didn't deeply research how this works in dkms. But given that a lot of distributions use dkms, and that those need to sign modules with some distribution-provided certificate, my best guess is that dkms handles this on its own, or through hooks.

flodavid: It deed interfere. It should be good now.
It now uses the standard `shim-signed` MOK. It can still be useful to help the user to enrol it if it is not already done though.
The password may be asked only on the first install for initial 
enrolment, but not for auto rebuilds.

hid-xpadneo/sign_and_install_generic.sh
flodavid: Section removed

Makefile
```diff
@@ -54,20 +56,17 @@ install: build
...
 
 sign:
-ifeq ($(NO_SKIP_SIGN), y)
-	@openssl req -new -x509 -newkey rsa:2048 -keyout MOK.priv -outform DER -out MOK.der -nodes -days 36500 -subj "/CN=Custom MOK/"
-	@mokutil --import MOK.der
-else
-	echo "Skipping key creation"
-endif
-	@$(KSRC)/scripts/sign-file sha256 MOK.priv MOK.der hid-xpadneo.ko
+	@sudo update-secureboot-policy --new-key
```
kakra: The Makefile should not call sudo, especially it should not do so hidden, and it should not run by surprise (e.g., you may have unlocked sudo before, and then it would run without asking for passwords). This also makes no sense because `modprobe` requires sudo but doesn't use it here. I'd suggest to rather add a guard `[ "$(id -u)" = "0"  ] || { echo >&2 "need to run with sudo"; exit 1; }`.

There are exceptions to this rule but this one is obviously not it.

sign_and_install.sh
```diff
@@ -14,28 +15,22 @@ then error "Please run as root"
...
-      if [ ! -f "${dkms_mok_dir}/MOK.pem" ]
-      then
-        info "Still no certificate for private key found. Generating one"
-        generate_certificate
-      fi
       install_xpadneo
     else
       warn "info the key exist but is not trusted yet, reimporting it because it does not hurt. You missing a reboot?"
-      import_trust
+      sudo update-secureboot-policy --enroll-key
```
kakra: Same as above... Please don't "secretly" sudo within a script. Ask the user to run it with sudo permissions, otherwise refuse to run. I have made some scripts in `lib` which you can source, and which check for permissions and initialize variables. You could move some of your functions there (creating a new lib file).

sign_and_install.sh
```diff
@@ -14,28 +15,22 @@ then error "Please run as root"
 fi
 
 dkms_mok_dir="${dkms_mok_dir:-/var/lib/shim-signed/mok}"
-mokutil_509key_pass=$(cat "${dkms_mok_dir}/.openssl_pass" 2> /dev/null)
-mokutil_509key_pass="${mokutil_509key_pass:-$(openssl rand -hex 6)}"
 
 command -v openssl > /dev/null 2>&1 || ( warn "missing openssl, you will not be able to create a new MOK key")
 command -v mokutil > /dev/null 2>&1 || ( warn "missing mokutil, you will not be able to sign the module")
-ls /var/lib/shim-signed > /dev/null 2>&1 || ( warn "missing shim-signed, you will not be able to sign the module with dkms")
+ls /var/lib/shim-signed > /dev/null 2>&1 || ( error "missing shim-signed, unable to enrol MOK")
```
kakra: Don't use `ls` to check for presence of a directory entry. You're expecting either a readable file or a directory. Use `[ -d /var/lib/shim-signed ] || error "missing ..."` or if you want to support symlinks, use `[ -d /var/lib/shim-signed/. ] || ...`. See `man test` for more options. Also, is it necessary to spawn `( subshells )` here?

sign_and_install.sh
```diff
@@ -47,68 +42,16 @@ function mokutil_setup () {
...
-        -outform DER \
-        -out "${dkms_mok_dir}/MOK.der" \
-        -days 31500 \
-        -subj "/CN=${name}/" \
-        -addext "extendedKeyUsage=codeSigning" || error "issue creating cert"
-
-    generate_certificate
-    import_trust
+    sudo update-secureboot-policy --new-key
+    sudo update-secureboot-policy --enroll-key
```
kakra: Same... Don't be sneaky... ;-)

sign_and_install.sh
```diff
@@ -119,7 +62,9 @@ function install_xpadneo () {
   fi
   source "$(dirname "$0")/install.sh"
 
-  ls /lib/modules/"$(uname -r)"/updates/
+  if [ ! -f /lib/modules/"$(uname -r)"/updates/dkms/hid-xpadneo.ko.zst ] && [ ! -f /lib/modules/"$(uname -r)"/updates/dkms/hid-xpadneo.ko.xz ]  && [ ! -f /lib/modules/"$(uname -r)"/updates/dkms/hid-xpadneo.ko ]
```
kakra: Try `modinfo hid-xpadneo | awk '/^Filename:/ { print $2 }'` instead. It tells you exactly which module file is present.

sign_and_install.sh
```diff
@@ -157,12 +102,14 @@ then
   exit 0
 else
   if mokutil --sb-state | grep -q "SecureBoot enabled"
-  then if ! grep -q "hid_xpadneo" /proc/modules && info "hid_xpadneo is loaded"
+  then if ! grep -q "hid_xpadneo" /proc/modules
```
kakra: This will fail on systems which linked xpadneo statically into the kernel. Check for `[ -d /sys/module/hid_xpadneo ]`, it will tell you if the driver is present in the kernel. OTOH, if xpadneo is statically linked into the kernel, you won't have to try signing it in the first place. ;-)

Makefile
```diff
@@ -54,20 +56,17 @@ install: build
 	$(DKMS) add hid-xpadneo
 
 uninstall: VERSION
-	$(DKMS) remove "hid-xpadneo/$(shell cat VERSION)" --all
+	$(DKMS) remove "hid-xpadneo/$(shell cat VERSION)" --all || true
```
kakra: Maybe `|| { echo >&2 "warning: dkms remove failed, proceeding anyways"; true; }` (or your warning helper)?

kakra (changes requested):
Good cleanup but there are some issues.

Makefile
kakra: ~On another note: This `Makefile` is to be used by package managers. We should not mess with `dkms` here. The docs explicitly state that it's up to the package maintainer to handle dkms.~

Ignore that last note... You just added `|| true`. Why?

Makefile
```diff
@@ -47,9 +56,17 @@ install: build
...
 	rm -f $(MODPROBE_CONFS:%=$(PREFIX)$(ETC_PREFIX)/modprobe.d/%)
 	rmdir --ignore-fail-on-non-empty -p $(PREFIX)$(ETC_PREFIX)/modprobe.d $(PREFIX)$(ETC_PREFIX)/udev/rules.d $(PREFIX)$(DOC_PREFIX)
+
+sign:
+	@sudo update-secureboot-policy --new-key
+	@sudo update-secureboot-policy --enroll-key
+	@$(KSRC)/scripts/sign-file sha256 $(SHIM_MOK_DIR)/MOK.priv $(SHIM_MOK_DIR)/MOK.der $(LIB_PREFIX)/hid-xpadneo.ko
+	modprobe hid-xpadneo
+
+sign-install: all install sign
```
kakra: No need to pull in `all` here, `install` already pulls `build`.

Makefile
```diff
@@ -47,9 +56,17 @@ install: build
...
 	rm -f $(DOCS:%=$(PREFIX)$(DOC_PREFIX)/%)
 	rm -f $(UDEV_RULES:%=$(PREFIX)$(ETC_PREFIX)/udev/rules.d/%)
 	rm -f $(MODPROBE_CONFS:%=$(PREFIX)$(ETC_PREFIX)/modprobe.d/%)
 	rmdir --ignore-fail-on-non-empty -p $(PREFIX)$(ETC_PREFIX)/modprobe.d $(PREFIX)$(ETC_PREFIX)/udev/rules.d $(PREFIX)$(DOC_PREFIX)
+
+sign:
+	@sudo update-secureboot-policy --new-key
+	@sudo update-secureboot-policy --enroll-key
+	@$(KSRC)/scripts/sign-file sha256 $(SHIM_MOK_DIR)/MOK.priv $(SHIM_MOK_DIR)/MOK.der $(LIB_PREFIX)/hid-xpadneo.ko
+	modprobe hid-xpadneo
```
kakra: The recipe says "sign" - don't modprobe. Especially, `modprobe` won't reload the module, so the outcome is not deterministic.

Makefile
```diff
@@ -47,9 +56,17 @@ install: build
...
 uninstall: VERSION
-	$(DKMS) remove "hid-xpadneo/$(shell cat VERSION)" --all
+	$(DKMS) remove "hid-xpadneo/$(shell cat VERSION)" --all || true
 	rm -Rf "$(PREFIX)/usr/src/hid-xpadneo-$(shell cat VERSION)"
 	rm -f $(DOCS:%=$(PREFIX)$(DOC_PREFIX)/%)
 	rm -f $(UDEV_RULES:%=$(PREFIX)$(ETC_PREFIX)/udev/rules.d/%)
 	rm -f $(MODPROBE_CONFS:%=$(PREFIX)$(ETC_PREFIX)/modprobe.d/%)
 	rmdir --ignore-fail-on-non-empty -p $(PREFIX)$(ETC_PREFIX)/modprobe.d $(PREFIX)$(ETC_PREFIX)/udev/rules.d $(PREFIX)$(DOC_PREFIX)
+
+sign:
```
kakra: You need `$(LIB_PREFIX)/hid-xpadneo.ko` here.

kakra: You also need `$(SHIM_MOK_DIR)/MOK.priv $(SHIM_MOK_DIR)/MOK.der`...

Makefile
```diff
@@ -47,9 +56,17 @@ install: build
...
+	@sudo update-secureboot-policy --new-key
+	@sudo update-secureboot-policy --enroll-key
+	@$(KSRC)/scripts/sign-file sha256 $(SHIM_MOK_DIR)/MOK.priv $(SHIM_MOK_DIR)/MOK.der $(LIB_PREFIX)/hid-xpadneo.ko
```
kakra: If you add the dependency for the sign recipe, you can use `$<` to reference the first dependency aka `$(LIB_PREFIX)/hid-xpadneo.ko`.

kakra (commented):
I wonder if signing with the `Makefile` is even worth it. It should be handle by dkms. If you need this for development, maybe move it to `hid-xpadneo/Makefile` instead? That file also allows usage of sudo for developer convenience. I'm using that to rebuild and reload my module during development like this:

```bash
make -C hid-xpadneo reinstall
```

Makefile
flodavid: Because if the module has already been removed, we has this error and the other commands are not run:
```
dkms remove "hid-xpadneo/v0.9-230-g73fd8cd-dirty" --all
Error! The module/version combo: hid-xpadneo-v0.9-230-g73fd8cd-dirty is not located in the DKMS tree.
make: *** [Makefile:55: uninstall] Error 3
```
I think adding a warning would be redundant.

kakra: Ah okay, dkms writes a proper message... Then maybe `|| echo "dkms error: proceeding anyways"` - but `true` is good for now.

sign_and_install.sh
flodavid: Done

sign_and_install.sh
flodavid: Fixed

sign_and_install.sh
flodavid: I created a mok.sh lib

Makefile
kakra: Well, I thought about this: This is a fix not specific to your feature proposal. So I will cherry pick it.

sign_and_install.sh
flodavid: Fixed

sign_and_install.sh
flodavid: I now look into `/sys/module/hid_xpadneo`. I will try to to add a check to see if the module should not be signed because it is statically linked into the kernel

kakra:
> I have a problem on my system though. I guess I have firmware 5.x since the module is not working without `uhid`

I'm not sure how we should properly handle that. It is expected that bluez loads uhid if it is needed. xpadneo is not responsible to do that, we are a HID user, not a HID provider. So I think you should report it to your distribution, maybe they should link uhid statically into the kernel, or add a dependency.

Until then, you could just run:
```bash
# always load uhid on boot (requires systemd)
echo "uhid" | sudo tee -a /etc/modules-load.d/99-local.conf

# OR: load uhid with xpadneo (untested)
echo "softdep hid-xpadneo pre: uhid" | sudo tee -a /etc/modprobe.d/99-xpadneo-local.conf

# ALTERNATIVE: load uhid with bluetooth (untested, probably preferred)
echo "softdep bluetooth pre: uhid" | sudo tee -a /etc/modprobe.d/99-bluetooth-uhid.conf
```

If uhid-related reports keep popping up, we might need to add the latter into xpadneo, tho, I still think, distributions should add that to the standard kernel if they provide bluez with BLE support (or make the bluez service load the module). BLE input devices cannot be used without uhid. Thus, there's no point in adding a softdep to each driver which in theory might maybe support a Bluetooth HID device...

**EDIT:** I added another alternative.

sign_and_install.sh
kakra: modinfo will return `(builtin)` on the filename line if it is builtin, e.g. I have btrfs built into the kernel:
```bash
# modinfo btrfs
name:           btrfs
filename:       (builtin)
...
```

But if a module is unknown, it will return an error:
```bash
# modinfo unknown || echo "exit code: $?"
modinfo: ERROR: Module unknown not found.
exit code: 1
```

Makefile
kakra: > I think adding a warning would be redundant.

I don't think so: If the make output shows an error, we should log that we willingly ignored that. That way, an auditor knows we detected the error but chose to ignore that. So it is not redundant by definition.

That said, I added your suggestion to my patch queue: https://github.com/atar-axis/xpadneo/pull/463

sign_and_install.sh
flodavid: Now that I think about it, I do not think I need to worry about that, since, as you said, there is no need to sign it if the module is statically linked. And I already skip the signing & install when the module is loaded (as it should be if statically linked I guess)

kakra:
> What's strange is that I have no vibration when the controller is connected, even though there is `ff_connect_notify=1`

This can happen if xpadneo needs to bind lazily to the device via udev rules - it may not have worked during the first connection, and then only on second connect, it will work (and that means, also `uhid` is present, because we need that to speak to the controller). If xpadneo is already loaded, it will bind immediately and automatically.

You should see a difference in dmesg for both tries - if you can still replicate it...

> The one who does is `softdep bluetooth` in "99-bluetooth-uhid.conf".

Which proves that the driver providing the HID bus has to load it - which is bluez. xpadneo is only a HID consumer. It doesn't even know about BT. Most likely, `bluez` user-space should probably trigger loading `uhid` as soon as an BLE input device connects, loading just a `bluetooth` module isn't technically the correct trigger to load `uhid`.

If you asked me, `uhid` should be built statically into the kernel by your distribution. It's a needed core driver, not optional.

sign_and_install.sh
flodavid: The `[ -d /sys/module/hid_xpadneo ]` version seems to be good

sign_and_install.sh
flodavid: I now check for modprobe errors and use `[ -d /sys/module/hid_xpadneo ]` instead

Makefile
flodavid: done

Makefile
flodavid: done

Makefile
flodavid: fixed

Makefile
flodavid: module loading moved to a new `load` step

Makefile
flodavid: Added a new step to create MOK using `update-secureboot-policy` if available. If the commands does not generate them in `/var/lib/shim-signed/mok` (which it should _I think_) users will have to override `MOK_DIR` though.

flodavid:
I rebased on master and fixed the conflict.
I also fixed styling errors raised by the workflow. There are some left, but it is either on files I did not modify, or bash scripts that are flagged as missing a  SPDX-License-Identifier, but other bash files do not have it as well.

flodavid:
I re-tested signing using the generic install and the  signing is still not made during the compilation by the kernel, but manually signing the module after its compilation is working well.

I added a second commit, that adds rules to the generic Makefile that installs/uninstalls the same files as the DKMS Makefile (docs, udev rules, ...). Tell me if you prefer not to have this or to make a separate PR.

Example output for generic install with signing (MOK was already created):
```shell
$ sudo ./sign_and_install_generic.sh
[info] Secure boot is enabled, signing is needed
[info] MOK Key /var/lib/shim-signed/mok/MOK.der found, already trusted.
[info] start install_xpadneo
make -C .. VERSION
make[1]: Entering directory '/home/flodavid/Dev/xpadneo'
{ [ -n "" ] && echo  || git describe --tags --dirty; } >VERSION
make[1]: Leaving directory '/home/flodavid/Dev/xpadneo'
make -C /lib/modules/6.17.0-23-generic/build INSTALL_MOD_DIR="kernel/drivers/hid" LD=ld.bfd M=/home/flodavid/Dev/xpadneo/hid-xpadneo/src VERSION="v0.11-pre-44-g9b1dbf5" modules
make[1]: Entering directory '/usr/src/linux-headers-6.17.0-23-generic'
make[2]: Entering directory '/home/flodavid/Dev/xpadneo/hid-xpadneo/src'
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: x86_64-linux-gnu-gcc-13 (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
  You are using:           gcc-13 (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
  CC [M]  xpadneo/consumer.o
  CC [M]  xpadneo/core.o
  CC [M]  xpadneo/debug.o
  CC [M]  xpadneo/device.o
  CC [M]  xpadneo/events.o
  CC [M]  xpadneo/keyboard.o
  CC [M]  xpadneo/mappings.o
  CC [M]  xpadneo/mouse.o
  CC [M]  xpadneo/power.o
  CC [M]  xpadneo/quirks.o
  CC [M]  xpadneo/rumble.o
  CC [M]  xpadneo/synthetic.o
  LD [M]  hid-xpadneo.o
  MODPOST Module.symvers
  LD [M]  hid-xpadneo.ko
  BTF [M] hid-xpadneo.ko
Skipping BTF generation for hid-xpadneo.ko due to unavailability of vmlinux
make[2]: Leaving directory '/home/flodavid/Dev/xpadneo/hid-xpadneo/src'
make[1]: Leaving directory '/usr/src/linux-headers-6.17.0-23-generic'
[ "0" = "0"  ] || { echo >&2 "needs to be run with sudo"; exit 1; }
make -C /lib/modules/6.17.0-23-generic/build INSTALL_MOD_DIR="kernel/drivers/hid" LD=ld.bfd M=/home/flodavid/Dev/xpadneo/hid-xpadneo/src VERSION="v0.11-pre-44-g9b1dbf5" modules_install
make[1]: Entering directory '/usr/src/linux-headers-6.17.0-23-generic'
make[2]: Entering directory '/home/flodavid/Dev/xpadneo/hid-xpadneo/src'
  INSTALL /lib/modules/6.17.0-23-generic/kernel/drivers/hid/hid-xpadneo.ko
  SIGN    /lib/modules/6.17.0-23-generic/kernel/drivers/hid/hid-xpadneo.ko
  DEPMOD  /lib/modules/6.17.0-23-generic
Warning: modules_install: missing 'System.map' file. Skipping depmod.
make[2]: Leaving directory '/home/flodavid/Dev/xpadneo/hid-xpadneo/src'
make[1]: Leaving directory '/usr/src/linux-headers-6.17.0-23-generic'
mkdir -p /etc/modprobe.d /etc/udev/rules.d /usr/share/doc/xpadneo
install -D -m 0644 -t /etc/modprobe.d etc-modprobe.d/xpadneo.conf
install -D -m 0644 -t /etc/udev/rules.d etc-udev-rules.d/60-xpadneo.rules etc-udev-rules.d/70-xpadneo-disable-hidraw.rules
install -D -m 0644 -t /usr/share/doc/xpadneo ../NEWS.md ../docs/3P-BUGS.md ../docs/BT_DONGLES.md ../docs/CONFIGURATION.md ../docs/DEBUGGING.md ../docs/FIRMWARE_FLASHING.md ../docs/PACKAGING.md ../docs/README.md ../docs/SDL.md ../docs/TESTING.md ../docs/TROUBLESHOOTING.md
udevadm control --reload
install -D -m 0644 -t /usr/share/metainfo ../io.github.atar_axis.xpadneo.metainfo.xml
make check_sudo
make[1]: Entering directory '/home/flodavid/Dev/xpadneo/hid-xpadneo'
[ "0" = "0"  ] || { echo >&2 "needs to be run with sudo"; exit 1; }
make[1]: Leaving directory '/home/flodavid/Dev/xpadneo/hid-xpadneo'
update-secureboot-policy --enroll-key | grep -q --invert-match 'No DKMS modules installed' \
  || mokutil --import "/var/lib/shim-signed/mok/MOK.der"
Nothing to do.
SKIP: /var/lib/shim-signed/mok/MOK.der is already enrolled
/lib/modules/6.17.0-23-generic/build/scripts/sign-file sha256 /var/lib/shim-signed/mok/MOK.priv /var/lib/shim-signed/mok/MOK.der /lib/modules/6.17.0-23-generic/kernel/drivers/hid/hid-xpadneo.ko
rmmod hid-xpadneo || echo "proceeding anyways"
rmmod: ERROR: Module hid_xpadneo is not currently loaded
proceeding anyways
depmod -a
modprobe hid-xpadneo 
rm ../VERSION
filename:       /lib/modules/6.17.0-23-generic/kernel/drivers/hid/hid-xpadneo.ko
softdep:        pre: uhid
version:        v0.11-pre-44-g9b1dbf5
description:    Linux kernel driver for Xbox ONE S+ gamepads (Bluetooth), including rumble
author:         Kai Krakow <kai@kaishome.de>
author:         Florian Dollinger <dollinger.florian@gmx.de>
license:        GPL
srcversion:     FF4D64EAE1FAFD708AF1F63
alias:          hid:b0005g*v0000045Ep00000B13
alias:          hid:b0005g*v0000045Ep00000B22
alias:          hid:b0005g*v0000045Ep00000B05
alias:          hid:b0005g*v0000045Ep00000B20
alias:          hid:b0005g*v0000045Ep000002FD
alias:          hid:b0005g*v0000045Ep000002E0
depends:        hid
name:           hid_xpadneo
retpoline:      Y
vermagic:       6.17.0-23-generic SMP preempt mod_unload modversions 
sig_id:         PKCS#7
signer:         REDACTED Secure Boot signature key (xpadneo)"
sig_key:        REDACTED
sig_hashalgo:   sha256
signature:      REDACTED
parm:           trigger_rumble_mode:(u8) Trigger rumble mode. 0: pressure, 2: disable. (byte)
parm:           rumble_attenuation:(u8) Attenuate the rumble strength: all[,triggers] 0 (none, full rumble) to 100 (max, no rumble). (array of byte)
parm:           ff_connect_notify:(bool) Connection notification using force feedback. 1: enable, 0: disable. (bool)
parm:           quirks:(string) Override or change device quirks, specify as: "MAC1{:,+,-}quirks1[,...16]", MAC format = 11:22:33:44:55:66, no pulse parameters = 1, no trigger rumble = 2, no motor masking = 4, use Linux button mappings = 16, use Nintendo mappings = 32, use Share button mappings = 64, reversed motor masking = 128, swapped motor masking = 256, apply no heuristics = 512 (array of charp)
parm:           disable_mouse:(bool) Disable mouse device permanently. 0: allow mouse, 1: disallow mouse. (bool)
parm:           gamepad_compliance:(bool) Adhere to Linux Gamepad Specification by using signed axis values. 1: enable, 0: disable. (bool)
parm:           disable_deadzones:(bool) Disable dead zone handling for raw processing by Wine/Proton, confuses joydev. 0: disable, 1: enable. (bool)
parm:           disable_shift_mode:(bool) Disable use Xbox logo button as shift. Will prohibit profile switching when enabled. 0: disable, 1: enable. (bool)
parm:           debug_descriptor:(bool) Dump unpatched HID descriptor to dmesg. 0: auto, 1: always. (bool)
parm:           debug_hid:(bool) Debug HID reports. 0: disable, 1: enable. (bool)
[info] finished install_xpadneo
[info] hid_xpadneo loaded!
[info] finished install_xpadneo
```

flodavid:
Can you take a look @kakra ?

PR #578, opened 2025-12-22
Support explicit paddle key usages for 6.17+

Lawstorant:
With fallback for older kernels

### DIFF
--- a/hid-xpadneo/src/hid-xpadneo.c
+++ b/hid-xpadneo/src/hid-xpadneo.c
@@ -165,7 +165,7 @@ static const struct usage_map xpadneo_usage_maps[] = {
 	USAGE_MAP(0xC0224, MAP_STATIC, EV_KEY, BTN_SELECT),
 
 	/* map special buttons without HID bitmaps, corrected in event handler */
-	USAGE_MAP(0xC0081, MAP_STATIC, EV_KEY, BTN_PADDLES(0)),	/* Four paddles */
+	USAGE_MAP(0xC0081, MAP_STATIC, EV_KEY, BTN_GRIPR), /* Four paddles */
 
 	/* hardware features handled at the raw report level */
 	USAGE_IGN(0xC0085),	/* Profile switcher */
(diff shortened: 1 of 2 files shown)


Lawstorant:
Afaik it's all been done already. @endrift did PR to SDL and even Steam already supports that. Even more, if steam is run on kernel 6.17+, paddles will be mangled/absent of they won't use the new usages.

I tried this and additional buttons showed up in steam immadietely.

kakra:
@Lawstorant I've changed your code slightly to use the same usage codes for older kernels and no longer use trigger happy codes. Are there any specific games I could test which explicitly support paddles? The Steam controller mapping detects the paddles as explicit buttons already so the SDL part should be fine. But we are mapping the controller to Xbox 360 so Steam doesn't show explicit paddle buttons (we could still remap them, and as long as Steam Input is turned off, Proton should report them, too, as the SDL layer reports them as paddles).

Lawstorant:
There aren't any games that could support paddles as the paddles aren't a game-facing feature on xbox/PC. They are only ever meant to map to something else so it's already a plus that we can have them as user-facing buttons on Linux. They could be mappable in Linux-native games.

I'm more concerned about:
> I've changed your code slightly to use the same usage codes for older kernels and no longer use trigger happy codes.

I didn't do that as a fickle. Steam actively checks if these usages are defined in the system and if not (running on kernel below 6.17) it will fall back to Trigger Happy usages. End result is that of you use the same codes for older kernels, steam won't detect paddles (I already tested that woth xone months ago).

With proper usages, steam does absolutely see the extended buttons even if mapped as X360 controller.

Actually, what's up with that? Why not leave it as Xbox One or series controller?

Lawstorant:
As you can see, Steam doesn't have any UI elements ready for extended buttons for Xbox 360 controller, but it absolutely does pick up the paddles:

Screenshot From 2025-12-27 18-20-14<br/>
 
 

Screenshot From 2025-12-27 18-21-03

Lawstorant:
Let me check again with `xone` and lts kernel

kakra:
> Let me check again with `xone` and lts kernel

Maybe xone does set keys in the bitmap it never uses? And so Steam detects it wrong?

I'm currently preparing v0.9.8. Maybe we can get this into the old branch before I release v0.10.

Lawstorant:
No it doesn't, I'd know, I maintain it currently

kakra:
Okay, then I'll take the commit from my queue... Thanks for testing.

kakra:
Pushed to master. Let me know if anything is wrong by opening a new bug or PR. Thanks for you efforts.

kakra:
Force pushed again to fix two small mistakes, see commit 81ab562f7da468ebebce001aef35e4fb12b852bd.

PR #577, opened 2025-12-18
Driver relicensing / rewrite tracking (kernel module scope)

kakra:
*This is a preparatory step and does not affect current releases.*

> Scope: Only code that is linked into the kernel module (hid-xpadneo/src/*).
> Goal: End state is GPL-2.0-only for all kernel driver code.
> Non-driver code (docs/examples/tools) may remain GPL-3.0-or-later.

## Purpose of this PR

This PR documents and prepares the ongoing effort to make the xpadneo kernel driver code compatible with the Linux
kernel’s GPL-2.0-only licensing requirements.

The goal is to eventually have all code that is linked into the kernel module (`hid-xpadneo/src/*`) licensed under
GPL-2.0-only, while allowing non-driver code (documentation, examples, tooling) to remain under GPL-3.0-or-later.

## What this PR does

- It introduces a tracking table that categorizes existing driver contributions by author, scope, and complexity.
- Based on that classification, we distinguish between:
  - **non-trivial driver logic**, where explicit relicensing consent is preferred, and
  - **trivial or low-effort changes**, which will be re-implemented during the ongoing refactoring instead of chasing
    consent.
- This PR does **not** change the license of existing code yet. It documents the current state and the intended
  migration path.

## About contributor tagging

Only contributors whose changes include non-trivial driver logic are explicitly tagged and asked for consent. This is
intentional.

Minor fixes (e.g. one-line bugfixes, formatting, small table entries, simple parameter gates) are planned to be
re-implemented as part of the refactor, as doing so is typically simpler and less intrusive than retroactively
collecting formal approvals.

If you contributed driver code and believe your contribution has been misclassified here, please feel free to comment —
corrections are welcome.

## Scope clarification

- This effort applies **only** to code that is linked into the kernel module.
- Documentation, descriptors, examples, and tooling are explicitly out of scope for GPL-2.0-only and may remain under
  GPL-3.0-or-later.

## Next steps

- Collect explicit consent from selected contributors where applicable.
- Continue modularization and clean-room re-implementation of driver components.
- Track replaced or removed legacy code in this table as the refactor progresses.

This approach is intended to be transparent, fair to contributors, and aligned with kernel licensing expectations.

## Request for confirmation

If you’re okay with it, please confirm here that your driver contributions may be relicensed under GPL-2.0-only as part
of the ongoing refactor.

This request applies only to contributions that are linked into the kernel module itself.

Simply post a reply with:

> I’m fine with relicensing my driver contributions under GPL-2.0-only.

[...]

PR #586, opened 2026-01-31
Add AppStream metainfo with hardware modalias

Michael-Lloyd:
Adds an AppStream metainfo file with `<modalias>` entries for all supported controllers.

This resolves `appstream-metadata-missing-modalias-provide` warnings from packaging tools like lintian when building distribution packages. The warning occurs because the udev rules use `TAG+="uaccess"` without corresponding AppStream metadata to advertise hardware support.

The metainfo also allows software centers (GNOME Software, KDE Discover) and tools like isenkram on Debian/Ubuntu to automatically suggest installing xpadneo when compatible hardware is detected.

References:
- [AppStream Driver Documentation](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Driver.html)
- [Debian AppStream Guidelines](https://wiki.debian.org/AppStream/Guidelines)

---

- [ ] Depends on: https://github.com/atar-axis/xpadneo/pull/577

### DIFF
--- a/xpadneo.metainfo.xml
+++ b/xpadneo.metainfo.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="driver">
+  <id>xpadneo</id>
+  <name>xpadneo</name>
+  <summary>Advanced Linux Driver for Xbox One Wireless Gamepad</summary>
+  <description>
+    <p>
+      Wireless driver for Xbox One S, Xbox Elite Series 2, and Xbox Series X|S
+      controllers connected via Bluetooth. Supports force feedback, rumble
+      effects, trigger force feedback, battery level indication, and
+      working paddle buttons.
+    </p>
+  </description>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0-only</project_license>
+  <url type="homepage">https://github.com/atar-axis/xpadneo</url>
+  <url type="bugtracker">https://github.com/atar-axis/xpadneo/issues</url>
+  <provides>
+    <!-- Xbox One S / X -->
+    <modalias>hid:b0005g*v0000045Ep000002E0*</modalias>
+    <modalias>hid:b0005g*v0000045Ep000002FD*</modalias>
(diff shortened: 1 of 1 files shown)


Michael-Lloyd:
> I think all commits should be squashed. Additionally, v0.10 will be dual-licensed - can we set both licenses then?

For dual licensing: yes, AppStream supports SPDX expressions. Once v0.10's licensing is set, you can use:

`<project_license>GPL-2.0-only AND GPL-3.0-or-later</project_license>`

I can update that now if you would like, or wait until re-licensing is finalized.

kakra:
@Michael-Lloyd I wonder how we should handle this.

You're placing the file in the `hid-xpadneo` folder which will contain the GPL-2.0-only module sources (currently still GPL-3.0). So that folder and all sub folders will be strictly GPL-2.0-only after releasing v0.10.

But the remainder of the Git tree will stay on GPL-3.0 (docs, examples, unmaintained patches).

What's the scope of the metainfo? Does it apply to the sub tree only, or to the whole Git tree? Should we move it up to the project root and mention both licenses? Should we keep it in `hid-xpadneo` and still should mention both licenses? Or should we mention GPL-2.0-only because this subfolder will become GPL-2.0-only with the v0.10 release?

hid-xpadneo/xpadneo.metainfo.xml
```diff
@@ -0,0 +1,29 @@
...
+  <description>
+    <p>
+      Wireless driver for Xbox One S, Xbox Elite Series 2, and Xbox Series X|S
+      controllers connected via Bluetooth. Supports force feedback, rumble
+      effects, trigger force feedback, battery level indication, and
+      working paddle buttons.
+    </p>
+  </description>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-3.0</project_license>
```
kakra: Note for later: This will need to change after the license migration before merging into v0.10.

hid-xpadneo/xpadneo.metainfo.xml
kakra, line 1: Pending GPL-2.0-only migration before v0.10 tagging.

PR #600, opened 2026-03-24
add python script to generate xpadneo kernel patch from sources

GloriousEggroll:
Allows generating a kernel patch for xpadneo instead of installing via dkms or akmods.

Usage example for kernel 6.19.9 and xpadneo v0.10:
```
wget https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.19.9.tar.xz
tar -xf linux-6.19.9.tar.xz
git clone https://github.com/atar-axis/xpadneo
cd xpadneo
git checkout v0.10
cd ..
cp xpadneo/generate-xpadneo-patch.py .
./generate-xpadneo-patch.py 6.19.9
```
Result: xpadneo-kernel-integration.patch

Folder structure:
Parent folder->
  (Linux kernel source)
  (xpadneo source)
  generate-xpadneo-patch.py
  
  Update 3/29/26:
  Also generates udev rules and modprobe file needed for xpadneo.


generate-xpadneo-patch.py
```diff
@@ -149,6 +156,44 @@ def create_xpadneo_makefile(xpadneo_version):
...
+    return """# xpadneo - Xbox Wireless Controller driver udev rules
+# This file provides proper permissions and access for xpadneo controllers
+
+# Rebind driver to xpadneo
+ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", KERNEL=="0005:045E:*", KERNEL=="*:02FD.*|*:02E0.*|*:0B05.*|*:0B13.*|*:0B20.*|*:0B22.*", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
+
+# Tag xpadneo devices for access in the user session
+# MODE="0664" allows user read/write access for proper Steam Input detection
+# LIBINPUT_IGNORE_DEVICE prevents desktop environments from using controller as mouse
+ACTION!="remove", DRIVERS=="xpadneo", SUBSYSTEM=="input", ENV{ID_INPUT_JOYSTICK}=="1", TAG+="uaccess", MODE="0664", ENV{LIBINPUT_IGNORE_DEVICE}="1"
```
kakra: Do we want to ignore it as a mouse unconditionally? The purpose of mouse mode is controlling the desktop environment to launch games. I recently discovered that KDE Plasma has its own controller-to-mouse mode but cursor movement is pretty bad. There's a module option to disable mouse mode - it should be sufficient for desktop environments, isn't it? If not, we may still be able to query the mod param from the udev rule, so users who want mouse mode can have mouse mode.

generate-xpadneo-patch.py
```diff
@@ -149,6 +156,44 @@ def create_xpadneo_makefile(xpadneo_version):
 	synthetic.o
 """
 
+def create_udev_rules():
+    """Generate xpadneo udev rules content"""
+    return """# xpadneo - Xbox Wireless Controller driver udev rules
+# This file provides proper permissions and access for xpadneo controllers
+
+# Rebind driver to xpadneo
+ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", KERNEL=="0005:045E:*", KERNEL=="*:02FD.*|*:02E0.*|*:0B05.*|*:0B13.*|*:0B20.*|*:0B22.*", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
```
kakra: This one isn't needed if the module is baked into the kernel (given that the supported PIDs/VIDs are patched out of hid-microsoft).

kakra (commented):
Just a quick first review without having it tried for my own kernel patches yet.

kakra:
@GloriousEggroll I've reworked the patch generator to better fit my workflow, and to reuse files from the source tree instead of hard-coding them in the script. Please review and let me know if this works for you, so I can merge it into xpadneo.

If ready, I'll merge the commits in one squashed commit adding your co-authored-by.

I'm not very fluent in python so I used multiple AI agents to change and review the changes, including some manual cleanups.

Changes:

- The script has been moved to misc.
- The script now expects a git repo as the kernel source, and optionally a second argument for the kernel version.
- Generated output will be written into dedicated output directories.
- Instead of unpacking/copying kernel/xpadneo sources, the script now uses temporary git worktrees.
- Instead of using diff, the script now uses a temporary git commit and exports the generated commit.

You can now run the script from the xpadneo directory without providing a Linux source tree at a predefined location, instead an existing Linux git clone is needed and can be placed anywhere in the file system.

generate-xpadneo-patch.py
kakra: I took that for now and merged it into master. I didn't verify yet how it affects mouse mode but I'm pretty sure it wouldn't affect mouse mode. So this is probably a good change.

Issue #549, opened 2025-07-01
XBE2 recognized as 360 controller in Steam and everywhere else


Epicness7131:
I am on Fedora 42, the driver works amazing and fully works, however, there is one issue, the controller is universally recognized a xbox 360 controller rather then the XBE2, meaning the back paddles are not properly supported or shown when testing/using. When xpadneo is uninstalled the controller is recognized as XBE2 again, meaning this is a issue/quirk with the driver.  This is my only issue with the driver and everything else works fine, controller rumbles when connected battery is shown etc. If the controller is intended to be recognized as a 360 controller it should probably be made more clear in the documentation.

![Image](https://github.com/user-attachments/assets/b0ba0f69-3c03-4a13-bdda-6280294e627d)

![Image](https://github.com/user-attachments/assets/0581a066-06a2-4c24-87ba-fd54e57f28ee)

```
[ 2599.671255] xpadneo 0005:045E:0B22.000A: input,hidraw6: BLUETOOTH HID v11.30 Gamepad [Xbox Wireless Controller] on 34:6f:24:da:1f:00
[ 2599.671260] xpadneo 0005:045E:0B22.000A: controller quirks: 0x00000050
[ 2599.671262] xpadneo xpadneo_welcome_rumble start
[ 2599.671264] xpadneo 0005:045E:0B22.000A: testing weak motor: sustain 50ms release 50ms loop 2 wait 30ms
[ 2600.001520] xpadneo 0005:045E:0B22.000A: testing strong motor: sustain 50ms release 50ms loop 2 wait 30ms
[ 2600.331764] xpadneo 0005:045E:0B22.000A: testing trigger motors: sustain 50ms release 50ms loop 2 wait 30ms
[ 2600.662007] xpadneo xpadneo_welcome_rumble took 991ms
[ 2600.662014] xpadneo 0005:045E:0B22.000A: Xbox Wireless Controller [f4:6a:d7:44:67:cf] connected
[ 2600.834553] xpadneo 0005:045E:0B22.000A: mapping profiles detected
[ 2600.834558] xpadneo 0005:045E:0B22.000A: switching profile to 1
[ 2824.121217] xpadneo 0005:045E:0B22.000A: switching profile to 2
[ 2825.599214] xpadneo 0005:045E:0B22.000A: switching profile to 3
[ 2825.921472] xpadneo 0005:045E:0B22.000A: switching profile to 0
[...]

PR #603, opened 2026-03-29
WIP: Disable pid spoofing to allow steam input support, add module option to enable spoofing on older SDL2, update udev rules

### LINKED ISSUE #423: Xbox Wireless Controller is seen as Xbox 360 controller and cannot use share button
## Version of xpadneo

0.9.5

## Controller Model

- [ ] Xbox One S controller
- [ ] Xbox Elite 2 controller
- [X] Xbox Series X|S controller
- [ ] Other:

### Connection mode

- [X] Bluetooth connection
- [ ] USB cable (not yet supported)
- [ ] Xbox Dongle connection (not yet supported)

## Installed Software

- [ ] Anti-Micro (may affect button mappings)
- [ ] OpenRGB (may mess up mappings and rumble stability)
- [X] Steam Input (enabled by default via Steam Desktop client)
- [ ] Steam Link (usually via Raspberry Pi or other micro computers)
- [ ] devices with QMK firmware (may affect udev rules, similar to OpenRGB)
- [ ] netstick (shares input devices via network similar to Steam Link)
- [X] xboxdrv (user-space gamepad driver)
- [ ] xone (kernel-space gamepad driver using the Xbox dongle or USB)
- [ ] xow (alternative driver using the Xbox dongle)

## Protocol Information

Please help us identify at which layer the problem can be found if you want
to report mapping errors or if the controller fails to be detected:

- [X] Steam Proton games are having issues
- [X] Steam Linux-native games are having issues
    - [ ] I don't use Steam or did not try
- [X] games running through Lutris, wine and/or Bottles are having issues
    - [ ] I don't use Lutris, Bottles, wine or did not try
- [X] Linux-native games are having issues
    - [ ] I don't use native games or did not try
- [X] Other software is having issues (describe software and issues below)
- [ ] Running `evtest` is showing issues (describe the issues below)
    - Keep in mind that `BTN_NORTH` and `BTN_WEST` are intentionally swapped
- [ ] Running `jstest` is showing issues (describe the issues below)
    - [ ] I don't have this tool or don't know how to use it
- [ ] Running `gamepad-tool` is showing issues (post console output below)
    - [ ] I don't have this tool

Please describe how it is failing below in the next sections.

## Severity / Impact

- [X] I've read the docs and the bug reporting instructions
- [X] I've applied the latest firmware update to the controller
- [X] I've tried disabling or running without above mentioned software
- [] It does not work at all
- [ ] It used to work in a previous version
- [ ] It mostly works but sometimes it doesn't
- [ ] I found a work-around
- [X] I probably didn't figure it all out but it's too early to give up
- [X] I don't know how to ...
- [ ] It's too complicated
- [X] Fantastic work but ...
- [X] I can code and I want to help

## Describe the Bug

My Xbox Series X|S, connected to my computer with a Bluetooth dongle, works perfectly. With your module, I was even able to get back the rumble feature I lost by switching to Linux. However, I still have two problems which I think are linked:
- the controller is shown in `xboxdrv`, Steam and in the last `dmesg` line as an Xbox 360 controller ; thus cannot bind a manual input to the Share button
- the `showkey` utility does not receive any signal from the Share button (but does for all other non-directional buttons)

## Steps to Reproduce

Connect the controller, go in `xboxdrv` or Steam controller settings.
Exit, try to use the Share button in `showkey`.

## Expected Behavior

The module should expose the controller as a Xbox X|S controller.

## Screenshots / GIFs / Videos

Screenshot_20230503_182020

Screenshot_20230503_182116

## System Information

```console
# uname -a
Linux stagcrown 6.2.13-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 26 Apr 2023 20:50:14 +0000 x86_64 GNU/Linux
```

```console
# xxd -c20 -g1 /sys/module/hid_xpadneo/drivers/hid:xpadneo/0005:045E:*/report_descriptor | tee >(cksum)
00000000: 05 01 09 05 a1 01 85 01 09 01 a1 00 09 30 09 31 15 00 27 ff  .............0.1..'.
00000014: ff 00 00 95 02 75 10 81 02 c0 09 01 a1 00 09 33 09 34 15 00  .....u.........3.4..
00000028: 27 ff ff 00 00 95 02 75 10 81 02 c0 05 01 09 32 15 00 26 ff  '......u.......2..&.
0000003c: 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01 81 03 05 01 09  ...u.....%.u........
00000050: 35 15 00 26 ff 03 95 01 75 0a 81 02 15 00 25 00 75 06 95 01  5..&....u.....%.u...
00000064: 81 03 05 01 09 39 15 01 25 08 35 00 46 3b 01 66 14 00 75 04  .....9..%.5.F;.f..u.
00000078: 95 01 81 42 75 04 95 01 15 00 25 00 35 00 45 00 65 00 81 03  ...Bu.....%.5.E.e...
0000008c: 05 09 19 01 29 0c 15 00 25 01 75 01 95 0c 81 02 15 00 25 00  ....)...%.u.......%.
000000a0: 75 01 95 04 81 03 05 0c 0a b2 00 15 00 25 01 95 01 75 01 81  u............%...u..
000000b4: 02 15 00 25 00 75 07 95 01 81 03 05 0f 09 21 85 03 a1 02 09  ...%.u........!.....
000000c8: 97 15 00 25 01 75 04 95 01 91 02 15 00 25 00 75 04 95 01 91  ...%.u.......%.u....
000000dc: 03 09 70 15 00 25 64 75 08 95 04 91 02 09 50 66 01 10 55 0e  ..p..%du......Pf..U.
000000f0: 15 00 26 ff 00 75 08 95 01 91 02 09 a7 15 00 26 ff 00 75 08  ..&..u.........&..u.
00000104: 95 01 91 02 65 00 55 00 09 7c 15 00 26 ff 00 75 08 95 01 91  ....e.U..|..&..u....
00000118: 02 c0 c0                                                     ...
2986910699 1363
```

## Controller and Bluetooth Information

[xpadneo-btmon.txt](https://github.com/atar-axis/xpadneo/files/11384792/xpadneo-btmon.txt)
[xpadneo-dmesg.txt](h
[xpadneo-lsusb.txt](https://github.com/atar-axis/xpadneo/files/11384794/xpadneo-lsusb.txt)
ttps://github.com/atar-axis/xpadneo/files/11384793/xpadneo-dmesg.txt)

## Additional Context

Thank you for your awesome work ! I did not try to listen to the Share button input manually via some C++ code but I think there should be an easier way. I saw in the doc this specific button was bound to the KEY_RECORD keyboard input, but cannot find a way to get to it.

GloriousEggroll:

1. Disable PID spoofing by default for Steam
 Input compatibility

PID spoofing (pretending to be Xbox 360 controller 0x028E) prevented
Steam Input from properly detecting controller models. Modern SDL2 2.28+ (released 2023) handles
Xbox controller mappings correctly without needing this workaround.

Changes:
- Add enable_pid_spoof module parameter (default: disabled)
- Update documentation to reflect modern SDL2 compatibility
- Real PIDs now exposed for proper device detection by Steam Input

Users with legacy SDL2 (<2.28) games can enable spoofing via:
  modprobe hid_xpadneo enable_pid_spoof=1

Fixes: #423, #549

2. Enable hidraw access for Steam Input
 compatibility

The hidraw blocking rule (70-xpadneo-disable-hidraw.rules) prevented
Steam Input from accessing controller device information needed for
proper model detection and Elite paddle support.

Changes:
- Enable hidraw access by default in 60-xpadneo.rules (MODE="0660")
- Add LIBINPUT_IGNORE_DEVICE to prevent desktop mouse interference
- Make 70-xpadneo-disable-hidraw.rules optional (Steam Link only)
- Add install-steamlink-fix target for Steam Link users
- Update documentation with breaking changes for v0.11

With PID spoofing disabled (previous commit), Steam Input now needs
hidraw access to read the real controller PID and enable advanced
features like Elite paddle remapping.

Steam Link users can install the optional hidraw blocking rule via:
  sudo make install-steamlink-fix

Related: #423, #549


hid-xpadneo/src/xpadneo/core.c
```diff
@@ -189,27 +196,30 @@ static int core_probe(struct hid_device *hdev, const struct hid_device_id *id)
...
-	 * still detect our driver as the correct model. Currently this
-	 * maps all controllers to the same model.
 	 */
 	xdata->original_product = hdev->product;
 	xdata->original_version = hdev->version;
-	hdev->product = 0x028E;
-	hdev->version = 0x00001130;
 
-	if (hdev->product != xdata->original_product)
+	if (param_enable_pid_spoof) {
```
kakra, line 225: I think the if block should only set the product and version, the original if-block can stay in place: If the product/version values didn't change, nothing will be logged anyways.

Makefile
```diff
@@ -25,25 +26,27 @@ all: build
...
+	@echo "build                   Prepare the package for DKMS deployment"
+	@echo "clangd-lsp              Generate compile_commands.json for clangd (requires bear)"
+	@echo "install                 Install the package, documentation and DKMS source code"
+	@echo "uninstall               Uninstall the package, documentation and DKMS source code"
+	@echo "install-all             Install everything including optional targets"
 	@echo
 	@echo "Optional targets:"
-	@echo "install-metainfo  Install appstream metainfo"
+	@echo "install-metainfo        Install appstream metainfo"
+	@echo "install-steamlink-fix   Install hidraw blocking rule for Steam Link compatibility"
```
kakra: This makes it difficult for users to arbitrarily flip the flag. I think it should be possible to query the mod param from the udev rules instead.

Makefile
```diff
@@ -67,11 +70,18 @@ install: build
 install-metainfo:
 	install -D -m 0644 -t $(PREFIX)$(META_PREFIX) xpadneo.metainfo.xml
 
+install-steamlink-fix:
+	@echo "WARNING: Installing hidraw blocking rule for Steam Link"
+	@echo "         This will BREAK Steam Input! Only use for Steam Link setups."
+	install -D -m 0644 -t $(PREFIX)$(ETC_PREFIX)/udev/rules.d $(UDEV_RULES_OPTIONAL:%=hid-xpadneo/etc-udev-rules.d/%)
+	$(UDEVADM) control --reload
```
kakra: I'd rather see the udev reload referenced from one place only - but if we query the mod param from udev, this block becomes obsolete anyways.

kakra (changes requested):
I appreciate the changes, but as mentioned previously, I'd rather introduce it as a new option with no behavior change by default. Then, on or after v0.11, we can flip the default and mark the previous default as deprecated, then remove it completely in the following version because I planned to remove this old SDL work-around anyways. But I think we also need to handle the HID report patches better then so it matches the original PID/VID descriptors. Otherwise the reports may look broken to some hidraw users. As far as I know, SDL has workarounds in place for that as it looks at descripter properties and report lengths only without really comparing the PID. But there's not only SDL.

hid-xpadneo/src/xpadneo/device.c
```diff
@@ -12,10 +12,21 @@
...
 
 	xpadneo_debug_hid_report(hdev, r, len);
 
+	/*
+	 * Some BLE controllers (e.g. XBE2 0x0B22) require a GATT Write Request
+	 * (acknowledged) rather than the unacknowledged GATT Write Command that
+	 * hid_hw_output_report sends. Use hid_hw_raw_request with HID_OUTPUT_REPORT
+	 * and HID_REQ_SET_REPORT to force the acknowledged path for these devices.
+	 */
+	if (xdata->quirks & XPADNEO_QUIRK_RAW_RUMBLE)
```
kakra: I wonder if this solves observed firmware crashes in genuine Xbox controller firmware if we send bursts of rumble with intervals shorter than 10ms. If it does, we could use it as the default and get rid of the throttling.

kakra: @GloriousEggroll I think I will cherry-pick this and will try making it the default. Did you run any tests that confirm we need a quirk flag for this? If Xbox Controllers need ACKed rumble packets for better stability, we should probably make it the default, and could hopefully get rid of the complicated rumble throttling. What do you think?

I have all current original Microsoft controllers here for tests - all using BLE firmware 5.x, so I cannot test this with the old 4.x firmware. Maybe we should check the firmware version instead of a quirk flag?

The XBE2 controller was the first to introduce BLE, and it had a lot of stability problems we could only solve with aggressive rumble throttling - so I'd like to make your idea the default, without quirk flag, and remove throttling completely.

kakra (commented):
Let's discuss if I should cherry-pick some of the changes for further testing.

hid-xpadneo/src/xpadneo/device.c
GloriousEggroll: I havent tested consistent lengthy rumble/haptics, i just did a basic test to make sure it was working. i can probably get around to checking deeper into it later this week

kakra: Thanks, I'll just wanted to know if you already did some tests which would say "we cannot use raw rumble for all controllers". Then I'll test this with the controllers I have. If it works well and eliminates using throttling, we should probably also suggest a similar change for hid-microsoft, and maybe tell the SDL team they can remove their throttling code, too, that way. Although from tests I remember, GIP also needs such a throttling workaround (10ms).

kakra: @GloriousEggroll First tests look very promising. I've cherry picked your change minus the quirk flag, and ripped out our artificial throttling logic, and at least the XBE2 controller worked flawlessly although the tested games sent rumble commands too fast. This has been a very good find by you - thank you. I'll now test my commits a little more and then push them to master, later backporting to stable.

kakra:
@GloriousEggroll I've cherry picked some of your changes and improved some other areas of the code. You may want to rebase this PR.

hid-xpadneo/src/xpadneo/device.c
kakra: I've cherry picked and tested this. I dropped the quirk flag and instead added a heuristic detect if we can use HOGP - only that allows us to use raw HID reports.

hid-xpadneo/src/xpadneo/events.c
```diff
@@ -137,6 +137,11 @@ int xpadneo_events_raw_event(struct hid_device *hdev, struct hid_report *report,
 		data[14] = SWAP_BITS(data[14], 2, 3);
 	}
 
+	/* swap X (BTN_NORTH) with Y (BTN_WEST) for controllers with non-standard XY layout */
```
kakra, line 140: Does this swap the buttons to map properly to their name, or does this swap the buttons to properly map to their Xbox layout position? I found that people prefer layout by position, not by name, so even is such controllers have Y in the west, people prefer Y to be the north button.

hid-xpadneo/src/xpadneo/device.c
```diff
@@ -172,29 +187,48 @@ const __u8 *xpadneo_device_report_fixup(struct hid_device *hdev, __u8 *rdesc, un
...
-	if (hdev->product == 0x02E0 && *rsize >= 0xac &&
-	    rdesc[0xa6] == 0x05 && rdesc[0xa7] == 0x01 &&
-	    rdesc[0xa8] == 0x09 && rdesc[0xa9] == 0x80) {
-		hid_notice(hdev, "8BitDo Pro 2 detected, spoofing as Xbox One 1697 (0x02DD) to bypass SDL3 HIDAPI\n");
-		xdata->quirks |= XPADNEO_QUIRK_NO_GUIDE_BTN;
-		hdev->product = 0x02DD;
+	if (hdev->product == 0x02E0) {
+		int i;
+
+		for (i = 0; xpadneo_8bitdo_02e0_names[i]; i++) {
```
kakra: I wonder if we could use the `DEVICE_NAME_QUIRK` macro for this? Maybe add another macro to match both name **and** OUI, and add an optional OUI spoofing item to the struct?

kakra (commented):
Some comments on the recent commits...

hid-xpadneo/src/xpadneo/events.c
GloriousEggroll: this is the correct mapping for this device -- all of those 8bitdo devices I purchased and tested. (more on the way). The original mapping on this device registers x as y and y as x as if it were a nintendo controller, which is wrong. with this change x registers as btn_west and y registers as btn_north (xbox layout)

GloriousEggroll:
> @GloriousEggroll I've cherry picked some of your changes and improved some other areas of the code. You may want to rebase this PR.

I have a local rebase on top of master. I plan on force pushing once i have the rest of the changes in. The gulikit rumble changes still arent fixed yet so I'm working on that.

kakra:
I've added a few commits lately which further improve rumble stability. Depending on your current PR base, this shouldn't introduce any new conflicts.

hid-xpadneo/src/xpadneo/device.c
```diff
@@ -213,12 +213,40 @@ const __u8 *xpadneo_device_report_fixup(struct hid_device *hdev, __u8 *rdesc, un
...
+	 * └────────────┴──────────────────────────┴───────────────────────────┘
 	 */
 	if (hdev->product == 0x02E0) {
-		int i;
+		/*
+		 * GameSir controllers register themselves as "Xbox Wireless Controller"
+		 * with PID 02E0, making them indistinguishable from the real Xbox One S
+		 * by name. Use OUI to identify them and apply the 02DD spoof.
+		 */
+		if (strncasecmp("A0:5A:5D", hdev->uniq, 8) == 0) {
```
kakra, line 284: GameSir uses completely random OUIs, you cannot use OUI to identify those. See heuristic docs in the repo... I found I can somewhat reliably identify them by looking at the LAA bit and the descriptor length (see current quirks.c implementation). It's a complete mess, I'm not even sure if this passes radio regulations in some countries, but OTOH, those clone vendors rarely care.

GloriousEggroll: are you sure about this? is this based on user collected reports or do you own the devices?

GloriousEggroll: btw you've got notes for the Nova Lite family but from what I can find none of the Nova Lite devices connect as xbox controllers over bluetooth. These are the ones I could find:

GameSir X2 Bluetooth — its iPhone mode tells you to search for “Xbox Wireless Controller.”
GameSir T3s — its iPhone/MFi mode also pairs as “Xbox wireless Controller.”
GameSir Cyclone/Cyclone Pro — iPhone connection instructions say to search for “Xbox Wireless Controller.”
GameSir T4 Pro — iPhone connection instructions say to search for “Xbox Wireless Controller.”
GameSir T4 Mini — iOS instructions say to search for “Xbox Wireless Controller gamepad.”
GameSir G4 Pro — iPhone instructions say to search for “Xbox Wireless Controller.”

kakra: I've collected several logs from issue reports, and none of the encountered GameSir OUIs is in the official OUI registry, some even fall into other vendor OUIs, some OUIs have the multicast bit set. GuliKit has some officially registered OUIs, but I've found at least one report now which seems to use a random OUI which at least has the LAA bit set and thus is safe to use (LAA = locally administered address). But an OUI which has the multicast bit set is clearly wrong. I probably didn't collect all OUIs from reports because I only recorded this after I discovered that the OUIs are clearly random, not officially registered, collide with other vendors, or clearly have the wrong bits set, see here: https://github.com/atar-axis/xpadneo/blob/master/docs/heuristics/gamesir-nova.md

Microsoft has a lot OUIs they use, it's a futile effort trying to identify models by OUIs - it can only be best-effort in the best case, usually it's worse. I've started to create fingerprints of the descriptor length and crc16 instead, this currently lives in `debug.c` and is made in a way to log unknown checksums including a full descriptor dump so we can collect info. Combined with indicators like device names or OUI LAA/multicast bits set, this can make a useful fingerprinting heuristic. If we want to use that for quirk detection, we should probably move the fingerprinting out of the debug module into the quirks module (or maybe even a new fingerprinting module).

But trying to identify GameSir devices by OUI is not a path you want to take - trust me. ;-) For MS and GuliKit OUIs it *may* work but is still an effort. And in the case of GuliKit, I've now seen the first OUI which is seemingly random and not officially registered. But I'm not sure if this is an old model, and GuliKit stopped using random OUIs, or if GuliKit started using random OUIs.

My first approach was to use AND masks trying to identify GameSir - in the hope that only certain bits in the OUI are random. But that quickly proved unsuccessful as the first collisions with Microsoft OUIs appeared - when I realized that Microsoft registered a lot of new OUIs for recently sold controllers. But Microsoft at least uses officially registered OUIs. I stopped expecting that from China clones.

kakra: Some commits you may want to look into:

- https://github.com/atar-axis/xpadneo/commit/5f48724e903ad6cdcfe2d60e5fd1715d6bb72974
- https://github.com/atar-axis/xpadneo/commit/a8cf2cd49ad1a86e2cc1f0b7da2a7b37330f7966
- https://github.com/atar-axis/xpadneo/commit/cf0f415e7a2b052e09727f43440492599a9842e4
- https://github.com/atar-axis/xpadneo/commit/5a319c8d895c732b73ccd41cc0eb762211e4ea49
- https://github.com/atar-axis/xpadneo/commit/369264e74de6f3e7c306c4ece081b0054f6b4e92

GloriousEggroll: Yeah I looked into the nova device manuals, whoever reported them mustve been confused:

Super Nova Bluetooth modes:
PC (DS4), iOS/Switch, Android (NO XINPUT/XBOX)
https://cdn.shopify.com/s/files/1/2241/8433/files/GameSir-Super_Nova_Manual.pdf?v=1740990709

Nova Bluetooth modes:
iOS/Switch, Android (NO XINPUT/XBOX)
https://cdn.shopify.com/s/files/1/2241/8433/files/GameSir-Nova_manual.pdf?v=1710901509

Nova Lite Bluetooth modes:
iOS/Switch, Android (NO XINPUT/XBOX)
https://cdn.shopify.com/s/files/1/2241/8433/files/Manual-GameSir_Nova_Lite_EN.pdf?v=1743156237

Nova 2 Lite Bluetooth modes:
PC (DS4), iOS/Switch, Android (NO XINPUT/XBOX)
https://cdn.shopify.com/s/files/1/2241/8433/files/GameSir-Nova_2_Lite_manual.pdf?v=1747982416

hid-xpadneo/src/xpadneo/events.c
kakra: But Xbox controllers have button Y as the north button, and X as the west button - essentially we are swapping north and west in the driver so it properly maps to X and Y on the controller. From community feedback during the years of development, I found that people preferred using exactly that configuration, even if the north button is labelled as Y on their controllers. That's why the Nintendo quirk exists. My 8bitdo has X north, Y west, B south, A east. The quirk is for people who prefer the correct label over the "correct" position (according to what you know from Xbox controllers). The kernel itself seems to be completely off: It expects X north, Y west, but A south, B east.

I wonder why there are 8bitdo controllers now that only swap X and Y? Or is this really an error in the bit representation? Or does it now actually try to follow the Linux kernel idea of button positions?

hid-xpadneo/src/xpadneo/device.c
```diff
@@ -110,6 +110,41 @@ const __u8 *xpadneo_device_report_fixup(struct hid_device *hdev, __u8 *rdesc, un
...
+	 * item and two End Collections.  The HID parser hits the trailing 0x00,
+	 * treats it as an unknown Main item (tag 0x0), aborts, then reports
+	 * "unbalanced collection" because the two open collections (Application
+	 * and Logical) were never closed.
+	 *
+	 * Replace the dead "Logical Minimum(0)" item and trailing NUL with two
+	 * End Collections and shrink the descriptor by one byte.  The incomplete
+	 * Loop Count Output field is dropped, but xpadneo sends the rumble report
+	 * as a fixed raw struct so the HID descriptor definition of that field
+	 * is not required for correct operation.
```
kakra, line 139: I really don't like that solution but sure, we probably cannot easily expand the rdesc beyond its originally allocated size, so this is the logical way to go. Could you dump the broken descriptors into the docs directories? I never encountered these specific broken descriptors.

hid-xpadneo/src/xpadneo/device.c
```diff
@@ -143,19 +178,29 @@ const __u8 *xpadneo_device_report_fixup(struct hid_device *hdev, __u8 *rdesc, un
...
-			 * entry for this controller's BLE firmware layout, so it falls back
-			 * to generic sequential mapping which misreads the non-sequential
-			 * firmware button order. Enable LINUX_BUTTONS remapping so raw_event
-			 * reorders the button bits sequentially for both Steam Input and
-			 * native Linux.
+			 * Xbox controllers in Linux/Android BT mode expose a
+			 * non-sequential button layout: HID usage 0x90003 is an
+			 * empty hole (never asserted by the firmware); the physical
+			 * X button lives at usage 0x90004, Y at 0x90005, and so on.
+			 * This shifts every button from X onward by one slot relative
```
kakra, line 185: This is what I called the Android layout: These controllers with sparse bitmaps actually worked just fine in Android because the event symbols then exactly mapped the labels on the device - and that's probably what Microsoft wanted to do. The interesting observation has been: This HID layout has been different depending on whether you paired to a Linux or Windows Bluetooth stack. So Microsoft actually had especially Android in mind when doing that, not Linux. However, with some tricks (like sharing the pairing key between the Windows registry and bluez), you could trick the controller into Windows mode even while running Linux.

hid-xpadneo/src/xpadneo/core.c
```diff
@@ -246,7 +249,14 @@ static int core_probe(struct hid_device *hdev, const struct hid_device_id *id)
 		goto err_release_id;
 	}
 
-	ret = hid_hw_start(hdev, HID_CONNECT_DEFAULT);
+	/*
+	 * Expose Xbox controllers through the translated input devices only.
+	 * This suppresses /dev/hidraw nodes so user space is forced onto the
+	 * evdev path instead of bypassing xpadneo via HIDAPI/raw HID.
+	 */
+	ret = hid_hw_start(hdev, HID_CONNECT_HIDINPUT |
```
kakra, line 257: This will prevent usage of our hidraw C example which can be used do directly try some rumble. We should at least document that for now.

kakra:
@GloriousEggroll I've added a commit (https://github.com/kakra/xpadneo/commit/6e37098d76e7fed065b7f81880a41af33adc00fe) to merge the two new known descriptor checksums. Could you post the associated descriptor dumps so I can finalize my WIP commit? You should be able to easily dump the descriptor to dmesg using the `debug_descriptor=1` modparam (if you rebased to latest master recently). Thanks.

hid-xpadneo/src/xpadneo/rumble.c
```diff
@@ -276,17 +276,22 @@ static void rumble_test(char *which, const struct xpadneo_devdata *xdata,
...
+	 * stop the motors by sending a zero-magnitude report.
+	 *
+	 * Zero ALL magnitudes, not just the ones for the motor(s) under test.  The
+	 * pck struct retains the full set of magnitudes initialised in rumble_welcome
+	 * (weak=40, strong=20, right=10, left=10); if only the tested motor's field
+	 * is zeroed the others remain non-zero in the packet.  Controllers that do
+	 * not honour the enable mask (many clone devices) treat every non-zero
+	 * magnitude field as active and keep rumbling.  Zeroing everything is safe
+	 * for controllers that do honour the mask: they only update the motor(s)
+	 * indicated by pck.data.enable and ignore the rest.
```
kakra, line 288: This is solved this way to actually detect devices which do not properly honor the flags. If this doesn't work as it should, the logic above is probably not correct. It is expected that such controllers continue to rumble, to get an issue report, and than fix this device with proper quirk flags. It is more a behavior detection test than a pure welcome rumble, and exists for future plans.

H234598:
Additional real-world validation for the no-PID-spoofing direction in this PR:

- Nobara 44, kernel `7.2.0-202.nobara.fc44.x86_64`
- Elite Series 2 Bluetooth `045e:0b22`, firmware 5.17
- packaged xpadneo `v0.11-pre-63-g3acca9f`

With current xpadneo, the log showed PID `0x0B22` being changed to `0x028E`. Steam treated the controller as generic Xbox 360 and accepted no UI input after a power cycle, although raw evdev input was present. `ff_connect_notify=0` did not help.

I then unbound/unloaded xpadneo and let `hid-generic` expose the real `045e:0b22` PID. Steam immediately selected `SDL_JOYSTICK_HIDAPI_XBOX_ONE`, recognized `Xbox One Elite 2 Controller`, exposed all four paddles, and accepted input without rebooting.

I did not test this PR branch directly, but the controlled driver switch validates its intended modern-Steam behavior: preserving the real PID fixes classification and input on this setup. A module option that permits no-spoofing while retaining xpadneo's other features would avoid the current all-or-nothing blacklist workaround.

H234598:
Completed two-identical-controller testing adds an important boundary to my previous validation.

The no-spoofing behavior remains successful: with xpadneo absent and both Elite Series 2 controllers exposed by `hid-generic` as real `045e:0b22`, Steam recognized two `Xbox One Elite 2 Controller` devices, assigned separate indices/XInput slots, exposed all four paddles on both, and accepted simultaneous input.

However, one reconnect of controller 2 while controller 1 stayed connected left controller 2 shown as connected but unusable in Steam. Reconnecting controller 2 again did not fix it. Turning both controllers off and starting controller 2 first recovered it without restarting Steam or the OS; subsequent individual reconnects of either controller worked.

Steam reports invalid/missing unit serials for these devices and generates synthetic IDs with suffixes, while reconnect handling logs `modify disconnected mapping in-place`. I did not capture raw evdev events during the one failed interval.

So this PR's no-PID-spoofing direction fixes the real and independently observed Steam classification problem, but should not be expected by itself to resolve every symptom in #605. A separate Steam/SDL/BlueZ multi-device hotplug or identity-cache issue remains.

kakra:
> A separate Steam/SDL/BlueZ multi-device hotplug or identity-cache issue remains.

There is at least one device enumeration bug on cold-plug in Proton which seems to be resolved in latest Experimental. I'm not sure if there hides another bug. But since it is in Proton, the problem should only exist in games, not in the Steam client itself.

PID spoofing removal is targeted for v0.12 currently. Any optional spoofing disable switch will only surface very late just before v0.11 tagging, or early in the v0.12 cycle.

Issue #457, opened 2024-02-11
Xbox Series X|S controller recognized as 360 controller and no inputs detected in games


QuickArgentum:
## Version of xpadneo

0.9.5
## Controller Model

- [ ] Xbox One S controller
- [ ] Xbox Elite 2 controller
- [X] Xbox Series X|S controller
- [ ] Other:

### Connection mode

- [X] Bluetooth connection
- [ ] USB cable (not yet supported)
- [ ] Xbox Dongle connection (not yet supported)

## Installed Software

- [ ] Anti-Micro (may affect button mappings)
- [X] OpenRGB (may mess up mappings and rumble stability)
- [ ] Steam Input (enabled by default via Steam Desktop client)
- [ ] Steam Link (usually via Raspberry Pi or other micro computers)
- [X] devices with QMK firmware (may affect udev rules, similar to OpenRGB)
- [ ] netstick (shares input devices via network similar to Steam Link)
- [ ] xboxdrv (user-space gamepad driver)
- [ ] xone (kernel-space gamepad driver using the Xbox dongle or USB)
- [ ] xow (alternative driver using the Xbox dongle)

## Protocol Information

Please help us identify at which layer the problem can be found if you want
to report mapping errors or if the controller fails to be detected:

- [X] Steam Proton games are having issues
- [X] Steam Linux-native games are having issues
    - [ ] I don't use Steam or did not try
- [ ] games running through Lutris, wine and/or Bottles are having issues
    - [X] I don't use Lutris, Bottles, wine or did not try
- [X] Linux-native games are having issues
    - [ ] I don't use native games or did not try
- [ ] Other software is having issues (describe software and issues below)
- [ ] Running `evtest` is showing issues (describe the issues below)
    - Keep in mind that `BTN_NORTH` and `BTN_WEST` are intentionally swapped
- [ ] Running `jstest` is showing issues (describe the issues below)
    - [ ] I don't have this tool or don't know how to use it
- [X] Running `gamepad-tool` is showing issues (post console output below)
    - [ ] I don't have this tool

Please describe how it is failing below in the next sections.

## Severity / Impact

- [X] I've read the docs and the bug reporting instructions
- [X] I've applied the latest firmware update to the controller
- [X] I've tried disabling or running without above mentioned software
- [ ] It does not work at all
- [ ] It used to work in a previous version
- [ ] It mostly works but sometimes it doesn't
- [ ] I found a work-around
[...]

Issue #303, opened 2021-07-17
Installing xpadneo and using Xbox One Controller over Bluetooth flips X+Y buttons and Start+Select buttons stop functioning until uninstalled. Over USB is fine.


ipaqmaster:
## Important note

This is a continuation of my documented experience in this issue: https://github.com/ValveSoftware/steam-for-linux/issues/7531

Continuing template.. 

## Version of xpadneo

Experienced in:
  1. Installing directly from commit 45fef91771bc0d32087c8b66c230417d9249d996
  2. Installing from AUR entry: https://aur.archlinux.org/packages/xpadneo-dkms-git/ which currently provides version 0.9.r83.g45fef91 (Appears to be same as above)

## Controller Model

* [x] Xbox One S controller
* [ ] Xbox Elite 2 controller
* [ ] Xbox Series X|S controller
* [ ] Other:

### Connection mode

* [x] Bluetooth connection
* [ ] Xbox Dongle connection (not yet supported)
* [x] USB cable (not yet supported) [Issue does not happen when I use USB]

## Installed Software

* [x] Steam Input (enabled by default via Steam Desktop client)
* [ ] Steam Link (usually via Raspberry Pi or other micro computers)
* [ ] xow (alternative driver using the Xbox dongle)
* [ ] Anti-Micro (may affect button mappings)
* [ ] netstick (shares input devices via network similar to Steam Link)
* [ ] xboxdrv (user-space gamepad driver)

## Severity / Impact

* [x] I've read the docs and the bug reporting instructions
* [x] I've applied the latest firmware update to the controller
* [ ] It does not work at all
* [x] It used to work in a previous version
* [ ] It mostly works but sometimes it doesn't
* [ ] I found a work-around
* [x] I probably didn't figure it all out but it's too early to give up
* [ ] I don't know how to ... 
* [ ] It's too complicated
* [ ] Fantastic work but ... 
* [ ] I can code and I want to help

## Describe the Bug

As title. Installing xpadneo either directly from here or via an available AUR repo package for Archlinux and using my Xbox One Controller over Bluetooth flips the X+Y buttons in all titles. It seems my Start+Select buttons are also ignored. USB is fine but xpadneo claims it isn't supported yet, so using USB is likely just not using xpadneo.

This is only something I wanted to report as this all used to work just fine. Utilities like `jstest /dev/input/js0` clearly show the buttons responding so I feel like it's a Steam or xpadneo thing.

[...]

PR #633, opened 2026-09-02
udev: close race where Steam grabs hidraw before xpadneo rebind

### LINKED ISSUE #1: Wrong mapping (when switching from Windows to Linux)
Sometimes the controller is sending different mappings, we already know how they look - but not how we can distinguish at device probing.

We therefore added a lot of debugging infos - **whenever you recognize a wrong mapping, please send us your `dmesg` output!**

if you wanna help: https://github.com/atar-axis/xpadneo/issues/1#issuecomment-367636301

### LINKED ISSUE #2: add more distributions to taskfile
The taskfile is tested only on ARCH linux on it's current version.

Hyphaed:
## Contributor Declaration

I agree that my contribution is accepted under the project's existing license terms, and that contributions to the module source code are accepted under **GPL-2.0-only**.

This contribution was created with the help of AI, including prior research with AI chatbots and AI-assisted reviews or suggestions. I will **not** add commits with a `Co-authored-by` signature for an AI agent. I have personally verified, reviewed, and tested the changes, and I take full responsibility for them as the author of the commits using my own `Signed-off-by`.

---

## Summary

This fixes a race between **xpadneo**, **udev/logind**, and **Steam's HIDAPI handling** that can leave a `hidraw` device with a stale user ACL.

The result is particularly confusing: the device appears to have been successfully locked down by xpadneo (`mode 0000`), but an ACL granted earlier by logind still gives the desktop user read/write access. Steam can therefore open the raw HID device directly instead of using the evdev interface exposed by xpadneo.

When that happens, Steam selects its HIDAPI Xbox driver and the controller repeatedly disappears:

```text
Controller using HIDAPI driver, vid=0x045e, pid=0x028e
Controller device closed after hid_read failure
Controller PollState Changed from 1 to 0
```

Games using Steam Input then fail to see the controller reliably.

---

## What's happening

The existing `70-xpadneo-disable-hidraw.rules` rule prevents userspace access to the `hidraw` node once the device is bound to xpadneo. In other words, this is what normally keeps SDL/Steam on the evdev interface rather than accessing the controller through raw HID.

The problem is **when** that rule runs.

On the affected system, Steam's own `60-steam-input.rules` sees the hidraw node while the controller is still bound to `hid-generic`:

```text
KERNELS=="*045E:0B22*"
```

It adds the `uaccess` tag, and logind subsequently grants the desktop user an ACL on the device.

Only afterwards does `60-xpadneo.rules` unbind and rebind the controller to `xpadneo`. At that point the existing rule in `70-xpadneo-disable-hidraw.rules` matches:

```text
DRIVERS=="xpadneo"
```

and removes `uaccess`.

The important detail is that **removing `uaccess` does not make logind revoke an ACL that it has already created**.

That leaves the device looking like this:

```text
$ ls -l /dev/hidraw7
c---rw----+ 1 root root 240, 7 ...

$ getfacl /dev/hidraw7
user:<user>:rw-
```

So although the node has mode `0000`, the `+` indicates an ACL is still present, and that ACL continues to grant the user access.

Steam then opens the node directly, selects its HIDAPI Xbox 360 driver instead of evdev, and eventually hits the `hid_read` failure loop described above.

---

## Fix #1  Match the hidraw node before the ACL is created

The first commit adds a second udev rule that matches the hidraw node by **HID ID**, rather than relying solely on:

```text
DRIVERS=="xpadneo"
```

The HID ID remains stable across the `hid-generic` → `xpadneo` rebind. This allows the rule to remove `uaccess` during the same udev processing in which Steam's rule would otherwise grant it, preventing logind from creating the stale ACL in the first place.

The PID list mirrors the controller rebind patterns already present in `60-xpadneo.rules`.

The existing `DRIVERS=="xpadneo"` rule is intentionally retained as a fallback for devices that are already bound directly to xpadneo.

The desired end state is therefore:

```text
$ ls -l /dev/hidraw7
c---------- 1 root root 240, 7 ...
```

with no user ACL attached.

---

## Fix #2  Make hidraw opt-in

The udev fix closes the immediate race, but it still depends on the device being rebound before Steam's rule can grant access.

The second commit takes a more direct approach to the same class of problem: **don't create a hidraw node unless something explicitly needs one.**

A new module parameter is added:

```text
enable_hidraw=0
```

The default is intentionally **off**.

### Implementation

In `core.c`, `enable_hidraw` is introduced as a boolean module parameter. Unless it is enabled, `core_probe()` masks `HID_CONNECT_HIDRAW` out of `HID_CONNECT_DEFAULT`.

As a result, xpadneo no longer creates `/dev/hidraw*` for the controller by default, removing the opportunity for Steam/logind to race over that device in the first place.

Documentation has been updated accordingly:

* `CONFIGURATION.md` documents `enable_hidraw`.
* `README.md` and `TROUBLESHOOTING.md` no longer assume that hidraw is always available.
* `misc/examples/c_hidraw` remains supported; users of that example should load the module with `enable_hidraw=1`.

This second change alters the default behaviour of a semi-stable branch. Per the release workflow guidance in `CONTRIBUTING.md`, it may therefore make sense to land it on its own schedule rather than coupling it to the udev fix.

---

## Test Results

The complete fix was tested against a real **Xbox Wireless Controller over Bluetooth LE (PID `0x0B22`)** using native, non-Flatpak Steam.

* [x] Reproduced the stale ACL / `hid_read failure` loop.
* [x] Confirmed that a Steam Input game correctly receives controller input after the fix.
* [x] Applied the udev rule and cleared the previously stale ACL.
* [x] Confirmed that `/dev/hidraw*` remains `mode 0000` with **no ACL** across a connect/disconnect cycle.
* [x] Confirmed that Steam no longer logs `Controller using HIDAPI driver` or the associated `hid_read failure` loop.
* [x] With `enable_hidraw` unset, confirmed that xpadneo creates **no `/dev/hidraw*` node** for the controller.
* [x] Confirmed that `misc/examples/c_hidraw` continues to work when the module is loaded with `enable_hidraw=1`.

## In short

The underlying problem is not that xpadneo fails to disable hidraw, it does. The problem is that **Steam/logind can grant access before xpadneo gets the opportunity to disable it, and that ACL survives the later rebind**.

The first commit prevents that stale ACL from being created.

The second commit removes the race entirely by making hidraw **opt-in rather than opt-out**.


kakra:
Thanks. This looks useful. I don't want to affront you, but this also looks like you created this with AI. If you did, your agent didn't follow the contribution rules and did not use the PR template. Please use the PR template when submitting PRs.

Hyphaed:
> Thanks. This looks useful. I don't want to affront you, but this also looks like you created this with AI. If you did, your agent didn't follow the contribution rules and did not use the PR template. Please use the PR template when submitting PRs.

you are right @kakra 

thanks for accepting the contribution

kakra:
Without using the PR template, I cannot accept this contribution. But I'll accept the fix, I will just implement it differently.

Hyphaed:
Fair, and no offense taken. I've updated the PR + pushed more code (missed to push the second fix earlier).

Yes, AI-assisted (Claude Code) for the investigation and patch, reviewed and tested by me on real hardware. I've updated the PR description with the template checkboxes and added a `Signed-off-by` line to the commit per CONTRIBUTING.md's AI-assistance rule.

kakra:
`enable_hidraw` is an unrelated change, and it's not going into v0.11 release. Please drop that, it is part of the idea in #603, or rather, a contrary change is part of it. Also, the second commit self-references the PR. Additionally, this introduces a behavioral defaults change which must not go into the same PR that is meant as a fix because it needs to apply separately after a major version bump.

I'd rather see a change like this:
```diff
+# Mark devices supported by xpadneo
+ACTION=="bind", SUBSYSTEM=="hid", KERNEL=="0005:045E:*", KERNEL=="*:02FD.*|*:02E0.*|*:0B05.*|*:0B13.*|*:0B20.*|*:0B22.*", ENV{ID_XPADNEO_SUPPORTED}="1"
+ACTION=="bind", SUBSYSTEM=="hid", KERNEL=="0005:0B05:1ABD.*", ENV{ID_XPADNEO_SUPPORTED}="1"
+
 # Rebind driver to xpadneo
-ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", KERNEL=="0005:045E:*", KERNEL=="*:02FD.*|*:02E0.*|*:0B05.*|*:0B13.*|*:0B20.*|*:0B22.*", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
-ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", KERNEL=="0005:0B05:1ABD.*", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
+ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", ENV{ID_XPADNEO_SUPPORTED}=="1", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
```

This deduplicates the ID definitions. On removal, we can check `ID_XPADNEO_SUPPORTED` then.

kakra:
Additionally, your AI agent still did not follow the PR template, instead it rephrased the checkboxes into elaborate sentences. I'll fix this in the agent rules later. The checkbox must remain a human click, the agent must not be allowed to check those for the human, or rephrase them.

kakra:
Just to be clear: I cannot and don't intent to prevent AI contributions. But it needs to be transparent and follow rules. Thus, please wait until I updated the AI rules and made my mind up exactly before pushing an updated PR with updated texts. You can still prepare your local commits, tho. I'm seeing a big difference how my agent handles the repository, and how yours does, both are Claude. I need to find out why this happens. Also, I may rethink if the AI agent should be forced to enable the AI transparency checkbox, or if this must remain a human action. I'll also improve the agent rules to prevent walls of text which mainly repeat what the commit descriptions already say.

With your help, we can improve the situation. I hope that's okay for you.

The unrelated `enable_hidraw` flag should become another separate PR. I was already thinking about a similar idea but stepped away from it for now because it may become obsolete after we switch to post-v0.11 development. In any case, if you want something like this go in, do not provide defaults that change the current stable driver behavior.

Hyphaed:
> Just to be clear: I cannot and don't intent to prevent AI contributions. But it needs to be transparent and follow rules. Thus, please wait until I updated the AI rules and made my mind up exactly before pushing an updated PR with updated texts. You can still prepare your local commits, tho. I'm seeing a big difference how my agent handles the repository, and how yours does, both are Claude. I need to find out why this happens. Also, I may rethink if the AI agent should be forced to enable the AI transparency checkbox, or if this must remain a human action. I'll also improve the agent rules to prevent walls of text which mainly repeat what the commit descriptions already say.
> 
> With your help, we can improve the situation. I hope that's okay for you.
> 
> The unrelated `enable_hidraw` flag should become another separate PR. I was already thinking about a similar idea but stepped away from it for now because it may become obsolete after we switch to post-v0.11 development. In any case, if you want something like this go in, do not provide defaults that change the current stable driver behavior.

no offense taken

I will take in consideration for next commit, yet I hope it everything it is fixed in next release and no more commits will be needed

actually I was working on this and light came down... hence I do have some unpushed code

I was adding logics to provide better logging/debugging, based on other projects I've been working in

I was just adding everything in this PR as you noted "The unrelated enable_hidraw flag..."

At the moment I'm outside home... working trough laptop... code sitting at desktop PC

Hyphaed:
Got home again

 Pushed an update:
  1. `enable_hidraw` dropped from this PR. Parked on a separate branch with the default flipped to on, so stable behavior doesn't change; will open separately, or drop it if it's obsolete after the post-v0.11 switch.
2. IDs deduplicated into `ENV{ID_XPADNEO_SUPPORTED}`, per your diff. One deviation from what you  posted: I set it in a new `59-xpadneo-hidraw-preempt.rules` (sorts before `60-xpadneo.rules`) rather than in `60-xpadneo.rules` itself, because the hidraw `uaccess` tag is granted by the vendor `60-*.rules` that sort in between `59-` and `60-xpadneo.rules`, setting the ENV in `60-xpadneo.rules` would be too late for the rule it needs to preempt. Let me know if that's not what you had in mind.
  4. Template restored verbatim.

The branch itself is ready to look at whenever you get to it.
Will check AI-rules update whenever those are ready. 

Thanks for xpadneo, you've done a great work.

hid-xpadneo/etc-udev-rules.d/59-xpadneo-hidraw-preempt.rules
```diff
@@ -0,0 +1,14 @@
+# Single definition of the controller IDs xpadneo supports. Set here rather
+# than in 60-xpadneo.rules so it is available to the vendor 60-*.rules that
+# sort in between.
+ACTION!="remove", SUBSYSTEM=="hid",    KERNEL=="0005:045E:*", KERNEL=="*:02FD.*|*:02E0.*|*:0B05.*|*:0B13.*|*:0B20.*|*:0B22.*", ENV{ID_XPADNEO_SUPPORTED}="1"
+ACTION!="remove", SUBSYSTEM=="hid",    KERNEL=="0005:0B05:1ABD.*", ENV{ID_XPADNEO_SUPPORTED}="1"
+ACTION!="remove", SUBSYSTEM=="hidraw", KERNELS=="0005:045E:02E0.*|0005:045E:02FD.*|0005:045E:0B05.*|0005:045E:0B13.*|0005:045E:0B20.*|0005:045E:0B22.*", ENV{ID_XPADNEO_SUPPORTED}="1"
+ACTION!="remove", SUBSYSTEM=="hidraw", KERNELS=="0005:0B05:1ABD.*", ENV{ID_XPADNEO_SUPPORTED}="1"
```
kakra, line 7: I wonder if we should really check the subsystem here. It is already checked properly in all the rules that have the `ID_XPADNEO_SUPPORTED` condition.

hid-xpadneo/etc-udev-rules.d/59-xpadneo-hidraw-preempt.rules
```diff
@@ -0,0 +1,14 @@
...
+
+# 60-steam-input.rules and other vendor 60-*.rules tag this hidraw node
+# uaccess by hardware id, regardless of which driver is bound. Stripping the
+# tag later does not help: 73-seat-late.rules matched TAG=="uaccess" as added
+# earlier in the same event, runs the uaccess builtin, and logind writes the
+# ACL even though the tag ends up removed. Skip past those rules instead.
+ACTION!="remove", SUBSYSTEM=="hidraw", ENV{ID_XPADNEO_SUPPORTED}=="1", GOTO="xpadneo_hidraw_lockdown"
```
kakra, line 14: I like the idea of having an extra position 59 file, it likely solves a real problem: sorting before Steam rules. But I'm not sure if we can place a `goto` here. Even if it works, it's a logical break. What is the intention of this? Skipping over rules 60..69?

hid-xpadneo/etc-udev-rules.d/60-xpadneo.rules
```diff
@@ -1,6 +1,5 @@
 # Rebind driver to xpadneo
-ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", KERNEL=="0005:045E:*", KERNEL=="*:02FD.*|*:02E0.*|*:0B05.*|*:0B13.*|*:0B20.*|*:0B22.*", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
-ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", KERNEL=="0005:0B05:1ABD.*", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
+ACTION=="bind", SUBSYSTEM=="hid", DRIVER!="xpadneo", ENV{ID_XPADNEO_SUPPORTED}=="1", ATTR{driver/unbind}="%k", ATTR{[drivers/hid:xpadneo]bind}="%k"
```
kakra, line 2: Just verifying, no change request: This has a condition on subsystem "hid", thus no need to have the same condition in 59.

hid-xpadneo/etc-udev-rules.d/70-xpadneo-disable-hidraw.rules
```diff
@@ -1,3 +1,11 @@
+LABEL="xpadneo_hidraw_lockdown"
+
 #FIXME(issue-291) Work around Steamlink not properly detecting the mappings
 #FIXME(issue-457) Word around QMK overriding our hidraw rules
 ACTION!="remove", DRIVERS=="xpadneo", SUBSYSTEM=="hidraw", MODE:="0000", TAG-="uaccess"
+
+# Fallback for a device that reaches this file without taking the GOTO in
+# 59-xpadneo-hidraw-preempt.rules. ID_XPADNEO_SUPPORTED is stable across the
+# hid-generic -> xpadneo rebind, unlike DRIVERS=="xpadneo" above.
```
kakra, line 9: Please verify if this change (continuing from position 59) fixes issues 291 and 457. Maybe we can finally resolve this reliably with your fixes.

Makefile
```diff
@@ -3,7 +3,7 @@ DOC_PREFIX ?= /usr/share/doc/xpadneo
 META_PREFIX ?= /usr/share/metainfo
 
 MODPROBE_CONFS := xpadneo.conf
-UDEV_RULES := 60-xpadneo.rules 70-xpadneo-disable-hidraw.rules
+UDEV_RULES := 59-xpadneo-hidraw-preempt.rules 60-xpadneo.rules 70-xpadneo-disable-hidraw.rules
```
kakra, line 6: Not sure if we should call the file this way. Position 70 has a clear purpose and concern, but 59 does not: Is must be used together with 70 while position 60 can in theory work on it's own. It should thus probably rather be named "disable-hidraw", too. But TBH: I'm not very happy with the logic here. Position 70 has always been a rule meant to be dropped once hidraw is compatible with user-space expectations. Now it is paired with 59 which is fine per-se, but rule 60 is thus now indirectly coupled to both these rule files:

- we are introducing `ID_XPADNEO_SUPPORTED` which rule 60 depends no
- we are moving the device IDs over to a potentially discardable rule file
- we are skipping over other, potentially system, rules with a goto breaking visible logical flow

Sure, this looks like it resolves existing problems, but the whole PR misses an argument how it not breaks other rules and not introduces new problems. Instead, it just repeats what the commit already says but just in a more elaborate way.

kakra (changes requested):
I'm mostly fine with the style now but I think this still needs refinement. Especially around the logic flow, and how to resolve the idea how rule 70 (and now 59) was meant to be a simple file drop once hidraw compatibility lands.

But more important is an explanation of how this resolves a real problem (and possibly #291 and #457) **without** introducing new bugs or regressions.

**Update:** #291 and #457 are already closed, see below for staleness considerations.

hid-xpadneo/etc-udev-rules.d/70-xpadneo-disable-hidraw.rules
kakra: #303 is the followup to #291, and #303 has been closed due to inactivity. We should probably remove that reference in a separate earlier commit. I'm not sure if #457 should be closed but it is. So maybe a removal candidate, too.

Makefile
kakra, line 6: Not sure if we should call the file this way. Position 70 has a clear purpose and concern, but 59 does not: Is must be used together with 70 while position 60 can in theory work on it's own. It should thus probably rather be named "disable-hidraw", too. But TBH: I'm not very happy with the logic here. Position 70 has always been a rule meant to be dropped once hidraw is compatible with user-space expectations. Now it is paired with 59 which is fine per-se, but rule 60 is thus now indirectly coupled to both these rule files:

- we are introducing `ID_XPADNEO_SUPPORTED` which rule 60 depends no
- we are moving the device IDs over to a potentially discardable rule file
- we are skipping over other, potentially system, rules with a goto breaking visible logical flow

Sure, this looks like it resolves existing problems, but the whole PR misses an argument how it not breaks other rules and not introduces new problems.

kakra (changes requested):
I'm not sure what's wrong with Github but it kept one pending comment and one comment in draft, so if I submit again, at least one comment shows up twice. Sorry for the noise.

kakra:
Thanks for the detailed report and for following up on the contribution rules - this was a legitimate real-world bug, and your reproduction and analysis of the ACL race were spot on.

While reviewing this, I ended up implementing a fix for the same root cause independently, using a slightly different mechanism (see separate PR https://github.com/atar-axis/xpadneo/pull/635). Rather than adding a new `59-` rules file with a `GOTO`/`LABEL` pair spanning three files, it sets `ENV{ID_XPADNEO_SUPPORTED}` once on the `hid` parent during its very first `bind` event (regardless of driver), and imports it into the `hidraw` child event with `IMPORT{parent}`. That has a few advantages here specifically:

- `70-xpadneo-disable-hidraw.rules` is intentionally a single, self-contained file users can drop to re-enable hidraw access. A `GOTO` from a separate file into a `LABEL` inside it would make dropping that file break rule evaluation elsewhere, instead of just restoring the old (accessible) behavior.
- The device-ID match only needs to happen once, in the `hid` subsystem; it doesn't need to be repeated for `hidraw` as well.
- Because the marker comes from the parent device's own event (which always precedes the child's first `hidraw` event), it closes the race independent of where any given vendor's rule file happens to sort - it only needs to beat the systemd `uaccess` builtin, an ordering question that was already solved.

I'd like to go with that alternate implementation instead of merging this PR as-is - its commit already carries a `Closes:` trailer, so this PR will close automatically once that one lands on `master`. I want to say clearly first, though: the bug report and root-cause analysis here were correct and valuable. I'd really appreciate it if you could test the alternate fix on your original setup (Steam + affected controller) once it's up - you have the concrete reproduction, and your confirmation would give real confidence before it lands in a release.

Thanks again for digging into this and for working through the process changes with me.

